PR #23, opened 2020-04-20
Remove the duplicate code for integer encoding. And correct the integ…

Shylock-Hg:
…er range check.

For example:
If we store uint8_t i = 0xFF; to INT8 column, we'll get -1(int64_t) from rowreader later.
So I change it.


Shylock-Hg:
Stale

PR #89, opened 2020-07-13
Implement lookup for storage layer.

bright-starry-sky:
Lookup for storage layer.


bright-starry-sky:
Close this one. #121 will instead of this one.

PR #93, opened 2020-07-21
Refactor the index storage structure

bright-starry-sky:
1,  Reject STRING type field when create index. allows index to contain FIXED_STRING type field.
2, The nullable bit in index key, using '0' to indicate null value and using '1' indicate non-null value.


bright-starry-sky:
Close this one. #123 instead of this one.

PR #110, opened 2020-08-17
Delete DownloadHandler and IngestHandler

laura-ding:


laura-ding:
2.0 is also need, so close it.

PR #111, opened 2020-08-17
Tuning GetNeighbors Perf

critical27:
Better to review by each commit, some modification is scattered among lots of files.
Some major improvement:
1. Only retrieve srcId/type/rank/dstId in collectEdgeProps if needed
2. Only add FilterNode and AggregateNode if necessary.
3. Add FLAGS_enable_multi_versions, there is a string copy if we want to filter the edge of different versions.
4. Allocate RowReader from stack instead of heap, similar to vesoft-inc/nebula#2249. The difference is that we use RowReaderWrapper to wrap reader of v1 and v2.

The Benchmark will retrieve 1 tag and 1000 edges, almost the same as vesoft-inc/nebula#2305

After these tuning, the benchmark of 2.0 has ~20% improvement than 1.0.


panda-sheep (dismissed):
This pr is same  as pr #103,  has reviewed in pr #103.

Good job, excellent!

src/codec/test/RowReaderV1Test.cpp
```diff
@@ -71,7 +71,7 @@ TEST(RowReaderV1, headerInfo) {
 
     // Empty row, return illegal schema version
     SchemaWriter schema5;
-    auto reader2 = RowReader::getRowReader(&schema5,
+    auto reader2 = RowReaderWrapper::getRowReader(&schema5,
                                            folly::StringPiece(""));
```
darionyaphet, line 75: alignment ?

src/storage/admin/RebuildEdgeIndexProcessor.cpp
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+#include "kvstore/LogEncoder.h"
+#include "storage/StorageFlags.h"
+#include "storage/admin/RebuildEdgeIndexProcessor.h"
+
+namespace nebula {
+namespace storage {
+
+void RebuildEdgeIndexProcessor::process(const cpp2::RebuildIndexRequest& req) {
```
darionyaphet, line 14: This file have replaced by RebuildEdgeIndexTask?

critical27: Good catch, add by mistake, fixed

bright-starry-sky (approved):
LGTM.

panda-sheep (approved):

darionyaphet (approved):
Great Work~

src/storage/exec/GetNeighborsNode.h
```diff
@@ -162,23 +163,15 @@ class GetNeighborsSampleNode : public GetNeighborsNode {
 
             auto edgeType = std::get<0>(sample);
             auto val = std::get<1>(sample);
+            reader = RowReaderWrapper::getEdgePropReader(planContext_->env_->schemaMan_,
+                                                         planContext_->spaceId_,
```
darionyaphet, line 167: Seems we don't need reset any more ？

critical27: Good question. Previously we need `reset` is because we don't need to reallocate if `reader` is not `nullptr`. For now, since we allocate wrapper from stack, we don't need it anymore. However, there are several `reset` method in `RowReaderWrapper`, they are for convenience, see `StorageIterator.h`. The `reset` is almost same as `getReader`.

PR #120, opened 2020-08-19
Fix package configuration

yixinglu:
- Update docker expose ports
- Install common rpm for storage and meta docker image

### DIFF
--- a/docker/Dockerfile.metad
+++ b/docker/Dockerfile.metad
@@ -7,6 +7,7 @@ RUN cd /home/nebula/BUILD/package \
 
 FROM centos:7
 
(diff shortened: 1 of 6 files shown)


docker/Dockerfile.metad
```diff
@@ -7,6 +7,7 @@ RUN cd /home/nebula/BUILD/package \
 
 FROM centos:7
 
+COPY --from=builder /home/nebula/BUILD/build/cpack_output/nebula-*-common.rpm /usr/local/nebula/nebula-common.rpm
```
Shylock-Hg, line 10: Why need this?

laura-ding (approved):

docker/Dockerfile.metad
yixinglu: Some resource files have been packaged in common rpm.

Shylock-Hg (approved):
LGTM

PR #121, opened 2020-08-19
Implement lookup for storage layer. 

bright-starry-sky:
Because upstream git repo changed to public. re-submit this PR instead of #89 .
And addressed comments.


bright-starry-sky:
- Optimize code logic
- Reduced function parameter
- Fixed g++  compile error
- Fixed heap-use-after-free error

PR #127, opened 2020-08-24
Disable STRING type fields create index. allow FIXED_STRING type.

bright-starry-sky:
Because repo changed to public. reopen this PR instead of #93
Addressed comments.


bright-starry-sky:
The plan has changed to support indexes of string type. close this one.

PR #131, opened 2020-08-26
Support to save default value to expression

laura-ding:
depends on https://github.com/vesoft-inc/nebula-common/pull/211

### DIFF
--- a/src/meta/processors/schemaMan/SchemaUtil.cpp
+++ b/src/meta/processors/schemaMan/SchemaUtil.cpp
@@ -112,8 +112,8 @@ bool SchemaUtil::checkType(std::vector<cpp2::ColumnDef> &columns) {
                     if (value.getStr().size() > typeLen) {
                         const auto trimStr = value.getStr().substr(0, typeLen);
                         value.setStr(trimStr);
-                        ConstantExpression fixedVale(value);
-                        column.set_default_value(Expression::encode(fixedVale));
+                        ConstantExpression fixedValue(value);
+                        column.set_default_value(Expression::encode(fixedValue));
                     }
                     break;
                 }
--- a/src/meta/processors/schemaMan/SchemaUtil.cpp
+++ b/src/meta/processors/schemaMan/SchemaUtil.cpp
@@ -109,10 +109,10 @@ bool SchemaUtil::checkType(std::vector<cpp2::ColumnDef> &columns) {
                     }
                     auto& colType = column.get_type();
                     size_t typeLen = colType.__isset.type_length ? *colType.get_type_length() : 0;
-                    if (value->getStr().size() > typeLen) {
-                        const auto trimStr = value->getStr().substr(0, typeLen);
-                        value->setStr(trimStr);
-                        ConstantExpression fixedVale(*value);
+                    if (value.getStr().size() > typeLen) {
+                        const auto trimStr = value.getStr().substr(0, typeLen);
+                        value.setStr(trimStr);
+                        ConstantExpression fixedVale(value);
                         column.set_default_value(Expression::encode(fixedVale));
                     }
                     break;
--- a/src/meta/processors/schemaMan/SchemaUtil.cpp
+++ b/src/meta/processors/schemaMan/SchemaUtil.cpp
@@ -3,35 +3,45 @@
  * This source code is licensed under Apache 2.0 License,
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
+
+#include "common/expression/ConstantExpression.h"
+
 #include "meta/processors/schemaMan/SchemaUtil.h"
+#include "utils/DefaultValueContext.h"
 
 namespace nebula {
 namespace meta {
 bool SchemaUtil::checkType(std::vector<cpp2::ColumnDef> &columns) {
+    DefaultValueContext mContext;
     for (auto& column : columns) {
         if (column.__isset.default_value) {
             auto name = column.get_name();
-            auto* value = column.get_default_value();
+            auto defaultValueExpr = Expression::decode(*column.get_default_value());
+            if (defaultValueExpr == nullptr) {
+                LOG(ERROR) << "Wrong format default value expression for column name: " << name;
+                return false;
+            }
+            auto value = Expression::eval(defaultValueExpr.get(), mContext);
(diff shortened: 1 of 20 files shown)


src/meta/processors/schemaMan/SchemaUtil.cpp
```diff
@@ -70,59 +80,61 @@ bool SchemaUtil::checkType(std::vector<cpp2::ColumnDef> &columns) {
...
                     }
                     auto& colType = column.get_type();
                     size_t typeLen = colType.__isset.type_length ? *colType.get_type_length() : 0;
-                    if (value->getStr().size() > typeLen) {
-                        const auto trimStr = value->getStr().substr(0, typeLen);
-                        value->setStr(trimStr);
+                    if (value.getStr().size() > typeLen) {
+                        const auto trimStr = value.getStr().substr(0, typeLen);
+                        value.setStr(trimStr);
+                        ConstantExpression fixedVale(value);
```
yixinglu, line 115: typo: fixedValue?

src/mock/MockData.cpp
```diff
@@ -296,31 +297,77 @@ std::unordered_map<std::string, std::vector<Serve>> MockData::teamServes_ = team
 // Mock schema
 std::shared_ptr<meta::NebulaSchemaProvider> MockData::mockPlayerTagSchema(SchemaVer ver) {
     std::shared_ptr<meta::NebulaSchemaProvider> schema(new meta::NebulaSchemaProvider(ver));
-    schema->addField("name", meta::cpp2::PropertyType::STRING, 0, false, "");
+    schema->addField("name",
```
yixinglu, line 300: Maybe we can provide a overload function `addField` and construct the constant expression inside, use the proxy pattern like following:

```c++
void addField(xx, ..., value) {
  addField(xx,..., new ConstantExpression(value));
}
```

So you needn't to update so many cases.

laura-ding: I don't think that's necessary, because normally it wouldn't be a value, it would be an expression.

src/meta/processors/schemaMan/SchemaUtil.cpp
laura-ding: good catch

CPWstatic (approved):

yixinglu (dismissed):

yixinglu (approved):

src/utils/DefaultValueContext.h
```diff
@@ -0,0 +1,72 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
```
critical27, line 1: Move it under src/storage/context?

src/storage/context/StorageExpressionContext.cpp
```diff
@@ -24,7 +25,8 @@ Value StorageExpressionContext::readValue(const std::string& propName) const {
         // read null value
         auto nullType = value.getNull();
         if (nullType == NullType::UNKNOWN_PROP && field->hasDefault()) {
-            return field->defaultValue();
+            DefaultValueContext expCtx;
+            return Value(Expression::eval(field->defaultValue(), expCtx));
```
bright-starry-sky, line 29: A question, What is the purpose of using this DefaultValueContext? if the `hasDefault()`  just return ConstantExpression. so can we using `StorageExpressionContext` direct？because the `expCtx` is unused in eval method of ConstantExpression class.

src/utils/DefaultValueContext.h
laura-ding: `codec` use it, so I don't think it's better to move to Storage.

src/storage/context/StorageExpressionContext.cpp
laura-ding: I think StorageExpressionContext is too redundant to the default value.

src/storage/context/StorageExpressionContext.h
```diff
@@ -184,6 +184,7 @@ class StorageExpressionContext final : public ExpressionContext {
     std::vector<std::pair<std::string, Value::Type>>  indexCols_{};
 };
 
+
```
panda-sheep, line 187: Please restore

src/storage/exec/QueryUtils.h
```diff
@@ -7,12 +7,14 @@
...
-namespace storage {
-
 #include "common/base/Base.h"
+#include "common/expression/Expression.h"
 #include "storage/CommonUtils.h"
 #include "storage/query/QueryBaseProcessor.h"
+#include "utils/DefaultValueContext.h"
+
+namespace nebula {
+namespace storage {
```
panda-sheep, line 17: Good catch！

src/utils/DefaultValueContext.h
```diff
@@ -0,0 +1,72 @@
...
+        return Value::kEmpty;
+    }
+
+    const Value& getInputProp(const std::string&) const override {
+        LOG(FATAL) << "Not allowed to call";
+        return Value::kEmpty;
+    }
+
+    void setVar(const std::string&, Value) override {
+    }
```
panda-sheep, line 58: do nothing here ？
 
What  does DefaultValueContext  do?

src/storage/exec/UpdateNode.h
```diff
@@ -53,7 +53,7 @@ class UpdateNode  : public RelNode<T> {
     kvstore::ResultCode getDefaultOrNullValue(const meta::SchemaProviderIf::Field* field,
                                               const std::string& name) {
         if (field->hasDefault()) {
-            props_[name] = field->defaultValue();
+            props_[field->name()] = Expression::eval(field->defaultValue(), *expCtx_);
```
panda-sheep, line 56: In update, `expCtx_ `   is `StorageExpressionContext`*

Is it okay to use it like this here?

panda-sheep (commented):
Good job！

critical27 (dismissed):
Generally LGTM, known issue, `now` may be different in in-edge and out-edge, toss can handle it.

critical27 (approved):

bright-starry-sky (approved):
LGTM

critical27:
UT failed https://github.com/vesoft-inc/nebula-storage/pull/131/checks?check_run_id=1284397217, will be handled in another PR.

PR #136, opened 2020-09-03
Feature/collect tags

Shylock-Hg:
[NG-181] Add a reserved property `_tags` to get tags of one vertex which typed `List<string>`. To implement it I add a implicit rule for getProps RPC that the properties format like `_LABEL` is reserved properties that will ignore the tag in process.

### DIFF
--- a/src/storage/query/QueryBaseProcessor.inl
+++ b/src/storage/query/QueryBaseProcessor.inl
@@ -15,46 +15,63 @@ namespace storage {
 
 template<typename REQ, typename RESP>
 cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::handleVertexProps(
-        std::vector<cpp2::VertexProp>& vertexProps) {
+        std::vector<cpp2::VertexProp>& vertexProps,
+        const std::vector<cpp2::VertexProp>& dummyTags) {
     for (auto& vertexProp : vertexProps) {
         auto tagId = vertexProp.tag;
-        auto iter = tagContext_.schemas_.find(tagId);
-        if (iter == tagContext_.schemas_.end()) {
-            VLOG(1) << "Can't find spaceId " << spaceId_ << " tagId " << tagId;
-            return cpp2::ErrorCode::E_TAG_NOT_FOUND;
-        }
-        CHECK(!iter->second.empty());
-        const auto& tagSchema = iter->second.back();
-        auto tagName = this->env_->schemaMan_->toTagName(spaceId_, tagId);
-        if (!tagName.ok()) {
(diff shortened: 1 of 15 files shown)


src/storage/query/QueryBaseProcessor.inl
```diff
@@ -15,46 +15,63 @@ namespace storage {
...
-            return cpp2::ErrorCode::E_TAG_NOT_FOUND;
-        }
-        CHECK(!iter->second.empty());
-        const auto& tagSchema = iter->second.back();
-        auto tagName = this->env_->schemaMan_->toTagName(spaceId_, tagId);
-        if (!tagName.ok()) {
-            VLOG(1) << "Can't find spaceId " << spaceId_ << " tagId " << tagId;
-            return cpp2::ErrorCode::E_TAG_NOT_FOUND;
-        }
+        if (!completeReservedVertexProps(vertexProp.get_props())) {
```
critical27, line 22: According to NG-181, `_tag` is a special tag of vertex, why we use as a reserved prop? It confuses me.

Shylock-Hg: Not special tag but a special property.

critical27: But actually there is no property of vertex, IMO, this PR involves some special handling. I think when query engine need to count the number of tags/edgeTypes of a vertex, all we need to do is pass all known tags/edgeTypes to storage. In this way, we almost don't need to change code. Of course, we could support return sth `tagId` like what we did to `kType` in edge.

Shylock-Hg: Yes a little special.But this due to the _tags is not the real property as others, so we can't handle it as normal properties.The query engine need the tags and do some thing base on these.So better could get List of tag by vertex id directly.

Shylock-Hg: And according to the feature require we need implement it directly in GetProps and GetNeighbors(not implemented now), so that query engine could check the tags in the response.

critical27: No matter in `GetProps` or `GetNeighbors`, the logic I mentioned is the exactly same in query engine. If a list is really needed, storage could return the list according to request (for example, add a column). In this PR, only tag in `GetProp` is handled, I think more special handling will arise when we deal with edge and go, especially bidirect edges need be taken into consideration.

Shylock-Hg: Yes good catch, maybe 0 is the better reserved id.

Shylock-Hg:
Replaced by #155 and flat the _tags to properties _exist.

PR #137, opened 2020-09-09
return EMPTY instead of NULL in GetNeighbors and GetProp

critical27:
The problem of NULL is that when trying to get multiple tag/edge, it is hard to tell whether the value is really NULL or the vertex doesn't have given tag/edge. For example, `fetch prop on * "Duncan"`.

@nevermore3, all NULL has been changed to EMPTY, including the stat column and expr column in GetNeighbors.

### DIFF
--- a/src/storage/exec/GetNeighborsNode.h
+++ b/src/storage/exec/GetNeighborsNode.h
@@ -49,7 +49,7 @@ public:
         // vertexId is the first column
         row.emplace_back(vId);
         // second column is reserved for stat
-        row.emplace_back(NullType::__NULL__);
+        row.emplace_back(Value());
 
         auto tagResult = hashJoinNode_->result().getList();
         for (auto& value : tagResult.values) {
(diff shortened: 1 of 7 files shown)


src/storage/exec/GetNeighborsNode.h
```diff
@@ -103,7 +102,7 @@ class GetNeighborsNode : public QueryNode<VertexID> {
             }
 
             // add edge prop value to the target column
-            if (row[columnIdx].type() == Value::Type::NULLVALUE) {
+            if (row[columnIdx].type() == Value::Type::__EMPTY__) {
```
laura-ding, line 105: row[columnIdx].empty()

src/storage/exec/GetNeighborsNode.h
```diff
@@ -155,7 +154,7 @@ class GetNeighborsSampleNode : public GetNeighborsNode {
         for (auto& sample : samples) {
             auto columnIdx = std::get<4>(sample);
             // add edge prop value to the target column
-            if (row[columnIdx].type() == Value::Type::NULLVALUE) {
+            if (row[columnIdx].type() == Value::Type::__EMPTY__) {
```
laura-ding, line 157: ditto

laura-ding (approved):

src/storage/exec/GetNeighborsNode.h
critical27: 👍, fixed.

laura-ding (approved):

nevermore3 (approved):

PR #143, opened 2020-09-16
fix storage crash in GetNeighborsProcessor and GetPropProcessor

jievince:
when the actual vid length in `go statement` and `fetch prop statement` is greater than spaceVidLen_,  then storaged service will crash.


darionyaphet:
`spaceVidLen` is defined when space created. Actually I'm not sure why the length in Go and Fetch is greater than it defined.

Could support some example ?

jievince:
> `spaceVidLen` is defined when space created. Actually I'm not sure why the length in Go and Fetch is greater than it defined.
> 
> Could support some example ?

I know `spaceVidLen` is defined when space created, but users can use Go and Fetch with any vertex of any length, even the vertex didn't exist in space. It also might

src/storage/query/GetPropProcessor.cpp
```diff
@@ -38,6 +38,15 @@ void GetPropProcessor::process(const cpp2::GetPropRequest& req) {
             auto partId = partEntry.first;
             for (const auto& row : partEntry.second) {
                 auto vId = row.values[0].getStr();
+
+                if (!NebulaKeyUtils::isValidVidLen(spaceVidLen_, vId)) {
+                    LOG(ERROR) << "Space " << spaceId_ << ", vertex length invalid, "
+                            << " space vid len: " << spaceVidLen_ << ",  vid is " << vId;
```
darionyaphet, line 44: alignment ?

jievince: ok

darionyaphet (approved):
LGTM ~

critical27:
Did query engine check the space vid length? If not, why not check after parser?

dutor:
> Did query engine check the space vid length? If not, why not check after parser?

Agreed.

The length checking shall be performed at the first place. But given that the vertex ids might be generated at runtime, I think the StorageClient is the best place to do this.

dutor:
BTW. @jievince Could you elaberate the details about the crash, like backtrace? I think it should not crash even if the vid is illegal

jievince:
> > Did query engine check the space vid length? If not, why not check after parser?
> 
> Agreed.
> 
> The length checking should be performed at the first place. But given that the vertex ids might be generated at runtime, I think the StorageClient is the best place to do this.

I think the query engine forget checking the vertex length, it just checks the vertex type now.
```C++
            if (!SchemaUtil::isValidVid(vid, space_.spaceDesc.vid_type)) {
                std::stringstream ss;
                ss << "Vid should be a "
                   << meta::cpp2::_PropertyType_VALUES_TO_NAMES.at(space_.spaceDesc.vid_type);
                return Status::Error(ss.str());
            }
            starts.vids.emplace_back(std::move(vid));
            startVidList_->add(expr->clone().release());
        }
    }
```

```C++
bool SchemaUtil::isValidVid(const Value &value, meta::cpp2::PropertyType type) {
    auto vidType = propTypeToValueType(type);
    if ((vidType != Value::Type::STRING
            && vidType != Value::Type::INT)  // compatible with 1.0
            || value.type() != vidType) {
        return false;
    }
    return true;
}
```

jievince:
> BTW. @jievince Could you elaberate the details about the crash, like backtrace? I think it should not crash even if the vid is illegal

image
image

critical27:
> > BTW. @jievince Could you elaberate the details about the crash, like backtrace? I think it should not crash even if the vid is illegal
> 
> image
> image

If so, we'd better check vid length in query engine, WDYT? In storage, only `Add vertex/edge` has handled this for now, others have not, we could add them in all processors if necessary.

dutor (approved):
LGTM.

PR #150, opened 2020-09-22
Support scan in storage

critical27:
Based on https://github.com/vesoft-inc/nebula-common/pull/246. Support scan data from storage within specified partition. Almost same as what we did in 1.0, except that we don't need to return the encoded value, just return the decoded value instead.

Could just review the new code, the existing code has expired which is from 1.0.

### DIFF
--- a/src/storage/exec/QueryUtils.h
+++ b/src/storage/exec/QueryUtils.h
@@ -86,6 +86,22 @@ public:
         return Status::Error(folly::stringPrintf("Invalid property %s", prop.name_.c_str()));
     }
 
+    static Status collectPropsInValue(RowReader* reader,
+                                      const std::vector<PropContext>* props,
+                                      nebula::List& list) {
+        for (const auto& prop : *props) {
+            if (prop.returned_ && prop.propInKeyType_ == PropContext::PropInKeyType::NONE) {
+                VLOG(2) << "Collect prop " << prop.name_;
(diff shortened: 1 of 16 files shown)


src/storage/exec/QueryUtils.h
```diff
@@ -86,6 +86,22 @@ class QueryUtils final {
         return Status::Error(folly::stringPrintf("Invalid property %s", prop.name_.c_str()));
     }
 
+    static Status collectPropsInValue(RowReader* reader,
+                                      const std::vector<PropContext>* props,
+                                      nebula::List& list) {
+        for (const auto& prop : *props) {
+            if (prop.returned_ && prop.propInKeyType_ == PropContext::PropInKeyType::NONE) {
+                VLOG(2) << "Collect prop " << prop.name_;
+                auto value = QueryUtils::readValue(reader, prop.name_, prop.field_);
```
bright-starry-sky, line 95: Do we need to determine the default values?

critical27: Yeah, default values has been handled in `readValue`

bright-starry-sky: > Yeah, default values has been handled in `readValue`

Got, excellent.

bright-starry-sky (approved):
LGTM

liuyu85cn (approved):
LGTM

PR #154, opened 2020-10-09
Fixed rebuild index  bug

panda-sheep:
When one tag or one edge contains multiple indexes, wrong index data will be generated.


darionyaphet:
Hold on please .Maybe I will support more interface such as rebuild specified index and rebuild all index on a tag or edge

panda-sheep:
> interface
I do not think so
This is a defect, and incorrect index data is generated
And I will use it when rebuild index (including statistic index)

Of course, when the function is enhanced later, it can be improved

panda-sheep:
> Hold on please .Maybe I will support more interface such as rebuild specified index and rebuild all index on a tag or edge

I do not think so
This is a defect, and incorrect index data is generated
And I will use it when rebuild index (including statistic index)
Of course, when the function is enhanced later, it can be improved

critical27 (approved):
Good catch!

bright-starry-sky:
LGTM. Thanks for fixing this.

bright-starry-sky (approved):

PR #155, opened 2020-10-09
Feature/check tag exists

Shylock-Hg:
Add a reserved property `_exist` for tag, true if tag attached to vertex false if not.Both getProps and getNeighbors supported.

### DIFF
--- a/src/storage/CommonUtils.h
+++ b/src/storage/CommonUtils.h
@@ -104,6 +104,11 @@ public:
                                        RowReader* reader,
                                        const std::string& ttlCol,
                                        int64_t ttlDuration);
+
+    static bool isReservedProperty(const std::string& name) {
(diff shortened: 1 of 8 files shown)


critical27:
Generally LGTM, one little question, do we need this for fetch edge as well?

Shylock-Hg:
> Generally LGTM, one little question, do we need this for fetch edge as well?

I don't think so , the edge type is part of the edge quaternion

src/storage/CommonUtils.h
```diff
@@ -104,6 +104,11 @@ class CommonUtils final {
                                        RowReader* reader,
                                        const std::string& ttlCol,
                                        int64_t ttlDuration);
+
+    static bool isReservedProperty(const std::string& name) {
```
monadbobo, line 108: Maybe we should make a table with all the reserved properties?

PR #160, opened 2020-10-12
Don't return the not exist key as row.

Shylock-Hg:
See https://github.com/vesoft-inc/nebula-graph/issues/324
For the edges, the single row of result may contains multi-edges so filter it in FetchEdges plan.


critical27 (approved):

panda-sheep (approved):
Excellent!

bright-starry-sky (approved):
LGTM.

PR #165, opened 2020-10-14
Move the files instead of copying when ingest SST files

darionyaphet:


src/kvstore/RocksEngine.cpp
```diff
@@ -330,6 +330,7 @@ int32_t RocksEngine::totalPartsNum() {
 
 ResultCode RocksEngine::ingest(const std::vector<std::string>& files) {
     rocksdb::IngestExternalFileOptions options;
+    options.move_files = true;
```
liuyu85cn, line 333: how about use a gflag here?

monadbobo: > how about use a gflag here?

I agree with you, or add a parameter for ingest.

liuyu85cn (dismissed):

src/kvstore/RocksEngine.cpp
```diff
@@ -336,6 +339,9 @@ int32_t RocksEngine::totalPartsNum() {
 
 ResultCode RocksEngine::ingest(const std::vector<std::string>& files) {
     rocksdb::IngestExternalFileOptions options;
+    if (FLAGS_move_files) {
+        options.move_files = true;
+    }
```
bright-starry-sky, line 344: What's about `options.move_files = FLAGS_move_files` ?

darionyaphet: Good ~

bright-starry-sky (approved):
Good job. LGTM

liuyu85cn (approved):

PR #168, opened 2020-10-17
Index write benchmark

bright-starry-sky:
This PR instead of #126.
Addressed panda-sheep's comment.


bright-starry-sky:
Addressed comments; Resolved conflict.

src/storage/test/StorageIndexWriteBenchmark.cpp
```diff
@@ -1,21 +1,23 @@
...
-#include "codec/RowWriterV2.h"
+#include "mock/MockCluster.h"
+#include "mock/MockData.h"
 #include <folly/Benchmark.h>
+#include "common/interface/gen-cpp2/storage_types.h"
+#include "common/interface/gen-cpp2/common_types.h"
+#include "utils/NebulaKeyUtils.h"
+#include "mock/AdHocIndexManager.h"
+#include "mock/AdHocSchemaManager.h"
+
```
panda-sheep, line 20: A small suggestion, put the header files of the same directory together

src/storage/test/StorageIndexWriteBenchmark.cpp
```diff
@@ -24,85 +26,183 @@ DEFINE_string(root_data_path, "/tmp/IndexWritePref", "Engine data path");
...
+            break;
+        }
+        case IndexENV::INVALID_INDEX : {
+            indexMan->addTagIndex(spaceId, indexId, -1, mockTagIndexColumns());
+            break;
+        }
+        case IndexENV::MULITPLE_INDEX : {
+            indexMan->addTagIndex(spaceId, indexId, tagId, mockTagIndexColumns());
+            indexMan->addTagIndex(spaceId, indexId + 1, tagId, mockTagIndexColumns());
+            indexMan->addTagIndex(spaceId, indexId + 2, tagId, mockTagIndexColumns());
```
panda-sheep, line 105: I have a question:
three indexes are created here, but all three  indexes contain the same column,
In practice, such a case does not exist.  here only for test?

src/storage/test/StorageIndexWriteBenchmark.cpp
```diff
@@ -119,9 +219,9 @@ void insertVertices(bool withoutIndex) {
             }
```
panda-sheep, line 219: line 103 ~ line 119 
This part of the code can be proposed separately
Avoid code duplication and Can be reused

src/storage/test/StorageIndexWriteBenchmark.cpp
```diff
@@ -161,34 +261,37 @@ void insertVertices(bool withoutIndex) {
         }
```
panda-sheep, line 261: line 142 ~ line 161
This part of the code can be proposed separately
Avoid code duplication and Can be reused

src/storage/test/StorageIndexWriteBenchmark.cpp
bright-starry-sky: > I have a question:
> three indexes are created here, but all three indexes contain the same column,
> In practice, such a case does not exist. Here only for test?

Good question, this test case used for write speed of the index.Although the structure of the index is the same, what is really written is three copies of data.

src/storage/test/StorageIndexWriteBenchmark.cpp
```diff
@@ -414,60 +527,107 @@ unmatchIndex: Without match index, and through asyncAtomicOp.
...
 --total_vertices_size=2000000
 ============================================================================
 src/storage/test/StorageIndexWriteBenchmark.cpprelative  time/iter  iters/s
 ============================================================================
-withoutIndex                                                  3.00s  333.45m
-unmatchIndex                                                  4.72s  212.02m
-attachIndex                                                  40.56s   24.66m
-duplicateVerticesIndex                                       46.14s   21.67m
-multipleIndex                                               1.19min   14.00m
+withoutIndex                                                  7.06s  141.71m
```
darionyaphet, line 535: It looks like the performance has dropped by half ?

critical27 (approved):

PR #171, opened 2020-10-19
Normalize suffix for index scan

bright-starry-sky:
An additional problem was detected when change this one ： issue #170 .
 This new problem will be perfected by other PR.

Issue #166, opened 2020-10-15
Storage returns string with extra paddings to expression


Aiee:
When querying a vertex key, storage sends the string with padding, and this causes incorrect result for relational expressions like IN, STARTS/ENDS WITH

The query used in the example below is ```GO FROM 'Tony Parker' OVER like WHERE like._dst ENDS WITH 'Ginobili' YIELD $^.player.name``` which should return a non-empty set. But it fetches an empty set due to the issue stated above.

![Screen Shot 2020-10-15 at 4 02 13 PM](https://user-images.githubusercontent.com/18348405/96095897-d0f47e00-0f01-11eb-9d90-3975bf6e86b6.png)

Issue #161, opened 2020-10-13
[UT]log_command_test


Shylock-Hg:
https://github.com/vesoft-inc/nebula-storage/runs/1245292086?check_suite_focus=true

PR #172, opened 2020-10-19
Fix storage would return wrong int vid ending with '\0'

### LINKED ISSUE #170: VertextID normalize for int type.
Currently, VID allows both fixed_string and integer types.
we need to normalize the return value for MetaClient.

for example :

```
int64_t vid = 1, should be return "\001\000\000\000\000\000\000"
fixed_string vid = "a\000\000\000\000\000\000" length is 8 , should be return "a"
```

critical27:
Rely on https://github.com/vesoft-inc/nebula-common/pull/263.

When the vid is integer, it could end with '\0', we can't simply discard them. So we need to be aware of vid type. It involves as follows:
1. GetNeighbors/GetProp/Update
2. StorageExpressionContext
3. Index, (not done, wait https://github.com/vesoft-inc/nebula-storage/pull/171 merge at first)

Fix unstable ut in ci as well.

close #170, #166, #161

### DIFF
--- a/src/storage/BaseProcessor.h
+++ b/src/storage/BaseProcessor.h
@@ -61,7 +61,7 @@ protected:
         if (!vIdType.ok()) {
             return cpp2::ErrorCode::E_SPACE_NOT_FOUND;
         }
-        isIntId_ = vIdType.value() == meta::cpp2::PropertyType::INT64 ? true : false;
+        isIntId_ = (vIdType.value() == meta::cpp2::PropertyType::INT64);
 
         return cpp2::ErrorCode::SUCCEEDED;
     }
--- a/src/storage/BaseProcessor.h
+++ b/src/storage/BaseProcessor.h
@@ -56,6 +56,13 @@ protected:
             return cpp2::ErrorCode::E_SPACE_NOT_FOUND;
         }
         spaceVidLen_ = len.value();
+
+        auto vIdType = this->env_->schemaMan_->getSpaceVidType(spaceId);
+        if (!vIdType.ok()) {
+            return cpp2::ErrorCode::E_SPACE_NOT_FOUND;
+        }
+        isIntId_ = vIdType.value() == meta::cpp2::PropertyType::INT64 ? true : false;
+
         return cpp2::ErrorCode::SUCCEEDED;
     }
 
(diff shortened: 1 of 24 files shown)


src/storage/context/StorageExpressionContext.cpp
```diff
@@ -65,9 +66,11 @@ Value StorageExpressionContext::getEdgeProp(const std::string& edgeName,
             return Value::kNullValue;
         }
         if (prop == kSrc) {
-            return NebulaKeyUtils::getSrcId(vIdLen_, key_);
+            auto srcId = NebulaKeyUtils::getSrcId(vIdLen_, key_);
+            return isIntId_ ? srcId : srcId.subpiece(0, srcId.find_first_of('\0'));
```
laura-ding, line 70: user can insert vid such as "aa\0bb". so graphd need to check vid if has `\0` and return an error to the user.

critical27: Yeah, graph need to prevent '\0' for string id.

bright-starry-sky (approved):
LGTM. excellent

panda-sheep (dismissed):
Well done! Excellent!

src/storage/BaseProcessor.h
```diff
@@ -56,6 +56,13 @@ class BaseProcessor {
             return cpp2::ErrorCode::E_SPACE_NOT_FOUND;
         }
         spaceVidLen_ = len.value();
+
+        auto vIdType = this->env_->schemaMan_->getSpaceVidType(spaceId);
+        if (!vIdType.ok()) {
+            return cpp2::ErrorCode::E_SPACE_NOT_FOUND;
+        }
+        isIntId_ = vIdType.value() == meta::cpp2::PropertyType::INT64 ? true : false;
```
liuyu85cn, line 64: isIntId_ = vIdType.value() == meta::cpp2::PropertyType::INT64

critical27: My bad, good catch!

bright-starry-sky (approved):
LGTM.

panda-sheep (approved):

bright-starry-sky (approved):

liuyu85cn (approved):

PR #174, opened 2020-10-20
support get all vertices on tag, get all edges on edgetype

panda-sheep:
Support get all vertices on tag, get all edges on edgetype through the following statement：
```
lookup on all tags where tag == xxx
lookup on all edges where edge == xxx
```

note：
also support indexes without fields， for example:
```
create tag  player();
insert vertex player() value 11:();  
create tag index player_index on player();

create edge e1();
insert edge e1() value 11->13:(); 
create edge index e1_index on e1();
```

### DIFF
--- a/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
+++ b/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
@@ -15,12 +15,6 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
     const auto &indexName = req.get_index_name();
     auto &edgeName = req.get_edge_name();
     auto &fieldNames = req.get_fields();
-    if (fieldNames.empty()) {
-        LOG(ERROR) << "The index field of an edge type should not be empty.";
-        handleErrorCode(cpp2::ErrorCode::E_INVALID_PARM);
-        onFinished();
-        return;
-    }
 
     std::set<std::string> columnSet(fieldNames.begin(), fieldNames.end());
     if (fieldNames.size() != columnSet.size()) {
@@ -80,6 +74,17 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
             continue;
         }
 
+        if (fieldNames.size() == 0) {
+            if (item.get_fields().size() != 0) {
+                checkIter->next();
+                continue;
+            } else {
+                LOG(ERROR) << "Index " << item.get_index_name() << " has existed";
+                resp_.set_code(cpp2::ErrorCode::E_EXISTED);
+                onFinished();
+                return;
+            }
+        }
         if (checkIndexExist(fieldNames, item)) {
             resp_.set_code(cpp2::ErrorCode::E_EXISTED);
             onFinished();
--- a/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
+++ b/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
@@ -35,7 +35,7 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
     folly::SharedMutex::WriteHolder wHolder(LockUtils::edgeIndexLock());
     auto ret = getIndexID(space, indexName);
     if (ret.ok()) {
-        LOG(ERROR) << "Create Edge Index Failed: " << indexName << " have existed";
+        LOG(ERROR) << "Create Edge Index Failed: " << indexName << " has existed";
         if (req.get_if_not_exists()) {
             resp_.set_id(to(ret.value(), EntryType::INDEX));
             handleErrorCode(cpp2::ErrorCode::SUCCEEDED);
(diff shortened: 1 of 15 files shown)


panda-sheep:
The ci test failed because the common repo has merged the pr of the default value.

bright-starry-sky:
Nice job. could you resolve conflict?

panda-sheep:
> Nice job. could you resolve conflict?

Fixed

panda-sheep:
log_append_test and log_command_test are unstable cases,
It will be fixed later.

src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
```diff
@@ -80,6 +74,17 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
...
+            if (item.get_fields().size() != 0) {
+                checkIter->next();
+                continue;
+            } else {
+                LOG(ERROR) << "Index " << item.get_index_name() << " has existed";
+                resp_.set_code(cpp2::ErrorCode::E_EXISTED);
+                onFinished();
+                return;
+            }
+        }
```
bright-starry-sky, line 87: I think if any index exists, there is no need to create a new index for get all edges index. just use any one of them. right?

panda-sheep: Good question！
First of all, I totally agree with you。

If there is any index on the tag or edge, there is no need to create a new index.
If there is no index on the tag or edge, you can create an index without fields to statis the all vertices of the tag or  the all edges of  the edgetype.

src/storage/index/LookupBaseProcessor.inl
```diff
@@ -33,17 +33,41 @@ cpp2::ErrorCode LookupBaseProcessor<REQ, RESP>::requestCheck(const cpp2::LookupI
...
+        // If no filed in fields, only _vid can be returned for tag.
+        // For edge, only _src, _ranking, _dst can be returned
+        if (fields.size() == 0) {
+            //
+            if (contexts_.size() != 1 || yieldCols_.size() != 0 ||
+                needFilter || ctx.get_column_hints().size() != 0) {
+                return cpp2::ErrorCode::E_INVALID_OPERATION;
+            }
+            noProp_ = true;
+        }
```
critical27, line 61: @bright-starry-sky, would you help to review this piece of block? It supposes to make sure that the query looks like `lookup on all tags where tag == xxx`, but the implementation confuses me.

bright-starry-sky: > @bright-starry-sky, would you help to review this piece of block? It supposes to make sure that the query looks like `lookup on all tags where tag == xxx`, but the implementation confuses me.

I guess，when` lookup on all tags where tag == xxx` , the IndexQueryContext should be as below :
```
struct IndexQueryContext {
    1: common.IndexID           index_id,        ----> The actual index ID
    // filter is an encoded expression of where clause.
    // Used for secondary filtering from a result set
    2: binary                   filter,                       -----> null
    // There are two types of scan: 1, range scan; 2, match scan (prefix);
    // When the field size of index_id IndexItem is not zero, the columns_hints are not allowed
    //    to be empty, At least one index column must be hit.
    // When the field size of index_id IndexItem is zero, the columns_hints must be empty.
    3: list<IndexColumnHint>    column_hints,      -----> empty
}
```

So just index id needed. right ？ if yes , I think the logic is good for me. A small suggestion is put this check logic outside the for loop, because the contexts_.size() must be one when  `where tag == xxx`.

critical27: Agreed. Just check first context is more clear.

panda-sheep: > A small suggestion is put this check logic outside the for loop, because the contexts_.size() must be one when `where tag == xxx`.

I put it here on purpose.
The reason is that if an ordinary lookup query contains an index without fields, an error will be reported.

So here need to check each `IndexQueryContext`.

bright-starry-sky: > > A small suggestion is put this check logic outside the for loop, because the contexts_.size() must be one when `where tag == xxx`.
> 
> I put it here on purpose.
> The reason is that if an ordinary lookup query contains an index without fields, an error will be reported.
> 
> So here need to check each `IndexQueryContext`.

Aha, If it were me, I would be inclined to do so as bellow. just my own thoughts

```
if (contexts_.size() == 1 && contexts_[0].column_hints.empty()) {
    ...
} else {
   ...
}
```

panda-sheep: I got you.
Here  includes the following two meanings:
1) When there is only an index with no fields, meet specific conditions
2) When an ordinary lookup, this special index cannot be included.

So I think every element in contexts_ needs to be checked.

src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
critical27: > I think if any index exists, there is no need to create a new index for get all edges index. just use any one of them. right?

Hold on... According to the code, I don't think this is implemented. For example, I can keep creating index with no fields of different names.

critical27: My bad, forget previous comment. I mean if index exists, no need to create one more empty index.

panda-sheep: Yeah，Good catch！Fixed.

critical27 (approved):
LGTM

bright-starry-sky (approved):

PR #181, opened 2020-10-24
Support zone

darionyaphet:


src/meta/KVBasedClusterIdMan.h
```diff
@@ -106,16 +106,16 @@ class ClusterIdMan {
...
-                           });
+                          if (code != kvstore::ResultCode::SUCCEEDED) {
+                              LOG(ERROR) << "Put failed, error "
+                                         << static_cast<int32_t>(code);
+                              ret = false;
+                          } else {
+                              LOG(INFO) << "Put key " << key
+                                        << ", val " << clusterId;
+                          }
+                              baton.post();
```
critical27, line 117: alignment

src/meta/processors/zoneMan/AddGroupProcessor.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (!nebula::ok(groupIdRet)) {
+        LOG(ERROR) << "Create Group failed";
+        handleErrorCode(nebula::error(groupIdRet));
+        onFinished();
+        return;
+    }
+
+    std::vector<kvstore::KV> data;
+    auto groupId = nebula::value(groupIdRet);
+    data.emplace_back(MetaServiceUtils::indexGroupKey(groupName),
```
critical27, line 55: group id and zone id is only used to check whether name conflict, right?

src/meta/processors/zoneMan/AddGroupProcessor.cpp
```diff
@@ -0,0 +1,90 @@
...
+                      std::string(reinterpret_cast<const char*>(&groupId), sizeof(GroupID)));
+    data.emplace_back(MetaServiceUtils::groupKey(groupName),
+                      MetaServiceUtils::groupVal(zoneNames));
+
+    LOG(INFO) << "Create Group: " << groupName;
+    doSyncPutAndUpdate(std::move(data));
+}
+
+bool AddGroupProcessor::checkGroupRedundancy(std::vector<std::string> zones) {
```
critical27, line 64: `const std::vector<std::string>&`? BTW, do we need to check this, I think group is more like a alias of several zones. It is ok to have multiple same alias.

src/meta/processors/zoneMan/DropZoneProcessor.cpp
```diff
@@ -0,0 +1,31 @@
...
+        handleErrorCode(cpp2::ErrorCode::E_NOT_FOUND);
+        onFinished();
+        return;
+    }
+
+    std::vector<std::string> keys;
+    keys.emplace_back(MetaServiceUtils::indexZoneKey(zoneName));
+    keys.emplace_back(MetaServiceUtils::zoneKey(zoneName));
+    LOG(INFO) << "Drop Zone " << zoneName;
+    doSyncMultiRemoveAndUpdate(std::move(keys));
```
critical27, line 27: Need to check all hosts in zone is empty?

src/meta/processors/zoneMan/AddGroupProcessor.cpp
darionyaphet: Currently they are use to check duplicate but later maybe act as the internal representation

src/meta/processors/zoneMan/AddGroupProcessor.cpp
darionyaphet: I'm afraid the two group create with the same zones although the order is different. Such as create group_0 z-0,z-1,z-2  and create group_1 z-2,z-1,z-0

src/meta/processors/zoneMan/DropZoneProcessor.cpp
darionyaphet: maybe we should check whether the zone belong to any group and the group is related to a space?

src/meta/MetaServiceUtils.cpp
```diff
@@ -475,16 +477,36 @@ std::string MetaServiceUtils::indexEdgeKey(GraphSpaceID spaceId,
...
+       .append(reinterpret_cast<const char*>(&type), sizeof(type))
        .append(reinterpret_cast<const char*>(&spaceID), sizeof(GraphSpaceID))
        .append(indexName);
     return key;
 }
 
+std::string MetaServiceUtils::indexGroupKey(const std::string& name) {
+    EntryType type = EntryType::GROUP;
+    std::string key;
+    key.reserve(128);
```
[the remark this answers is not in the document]
darionyaphet: yep but this is unified with another index key

critical27 (approved):
Generally LGTM

liuyu85cn (approved):
LGTM

PR #186, opened 2020-10-28
Support create space with zone (#181)

darionyaphet:


src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -196,10 +233,29 @@ CreateSpaceProcessor::pickHostsWithZone(PartitionID partId,
             LOG(ERROR) << "Get zone " << zoneName << " failed";
             handleErrorCode(cpp2::ErrorCode::E_STORE_FAILURE);
             onFinished();
-            return pickedHosts;
+            return Status::Error(folly::stringPrintf("Get Zone %s failed", zoneName.c_str()));
         }
```
critical27, line 237: kv of zone could be read at first, no need to check several times in loop of part and replica

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -126,8 +126,23 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
...
+            LOG(ERROR) << "Get host loading failed";
+            handleErrorCode(cpp2::ErrorCode::E_INVALID_PARM);
+            onFinished();
+            return;
+        }
+
+        auto hostLoading = hostLoadingRet.value();
         for (auto partId = 1; partId <= partitionNum; partId++) {
-            auto partHosts = pickHostsWithZone(partId, zones, replicaFactor);
+            auto partHostsRet = pickHostsWithZone(partId, zones, replicaFactor, hostLoading);
```
critical27, line 139: Do we need to check size of zones is greater than replicaFactor?

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -174,20 +189,42 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
 CreateSpaceProcessor::pickHostsWithZone(PartitionID partId,
                                         const std::vector<std::string>& zones,
-                                        int32_t replicaFactor) {
+                                        int32_t replicaFactor,
+                                        std::unordered_map<HostAddr, int32_t>& loading) {
     auto startIndex = partId;
     std::vector<HostAddr> pickedHosts;
     for (int32_t i = 0; i < replicaFactor; i++) {
-        auto zoneName = zones[startIndex++ % zones.size()];
+        auto& zoneName = zones[startIndex++ % zones.size()];
```
critical27, line 221: Is this right? If I understand correctly, for example, a group contains 4 zones, and we keep creating space of replica 3, they will be always on first 3 zones.

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -174,20 +189,42 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
         auto zoneIdRet = getZoneId(zoneName);
```
critical27, line 222: kv of zone could be read at first, no need to check several times in loop of part and replica

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
critical27, line 221: Do we need to pick zone wit lowest loading?

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
darionyaphet: yep  I have check it in Line 122

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
darionyaphet: when I append more zone into this group I wish to use the free zone

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
darionyaphet: Good Point & remove the duplicate check

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
critical27: Exactly, but current code is more like a round-robin of zones, not based on zone loading.

src/meta/test/MetaClientTest.cpp
```diff
@@ -1176,70 +1352,70 @@ class TestListener : public MetaChangedListener {
...
-        // Test Drop Space
-        auto ret = client->dropSpace("default_space_1").get();
-        ASSERT_TRUE(ret.ok()) << ret.status();
-    }
-    sleep(FLAGS_heartbeat_interval_secs + 1);
-    ASSERT_EQ(1, listener->spaceNum);
-    ASSERT_EQ(9, listener->partNum);
-    client->unRegisterListener();
-}
+// TEST(MetaClientTest, DiffTest) {
```
critical27, line 1355: Diff is affected?

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -174,34 +200,97 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
+        }
+
+        loading[picked] += 1;
+        pickedHosts.emplace_back(toThriftHost(std::move(picked)));
+    }
+    return pickedHosts;
+}
+
+StatusOr<std::vector<std::string>>
+CreateSpaceProcessor::pickLightLoadZones(const std::vector<std::string>& zones,
```
critical27, line 257: IMO, one way to do it is use priority_queue, for each partition, just dequeue read first `replica` zones, then enqueue again.

critical27 (commented):
Avoid read zoneKey many times. Others generally LGTM

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
darionyaphet: I have thought about heap (priority_queue) but the zone size is very small and use sort maybe more readable

critical27: Yeah, the pros is that heap could be use several times, no need to get zones loading every time when add a new partition.

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -174,34 +200,97 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
+    }
+    std::vector<std::string> pickedZones;
+    for (const auto &pair : sortedMap) {
+        pickedZones.emplace_back(pair.second);
+        int32_t size = pickedZones.size();
+        if (size == replicaFactor) {
+            break;
+        }
+    }
+    return pickedZones;
```
bright-starry-sky, line 292: what purpose of `std::multimap<int32_t, std::string> sortedMap;` ?

src/meta/processors/zoneMan/AddZoneProcessor.cpp
```diff
@@ -14,20 +16,31 @@ void AddZoneProcessor::process(const cpp2::AddZoneReq& req) {
...
     std::set<HostAddr> nodeSet(nodes.begin(), nodes.end());
     if (nodes.size() != nodeSet.size()) {
-        LOG(ERROR) << "Conflict host found in the zone.";
+        LOG(ERROR) << "Conflict host found in the zone";
         handleErrorCode(cpp2::ErrorCode::E_CONFLICT);
         onFinished();
         return;
     }
 
+    auto activeHosts = ActiveHostsMan::getActiveHosts(kvstore_, FLAGS_heartbeat_interval_secs * 2);
```
bright-starry-sky, line 33: All hosts ? or all storage hosts?

bright-starry-sky: > All hosts ? or all storage hosts?

That ok. the default host rule is storage.

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -174,34 +200,97 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
-        auto zoneIdRet = getZoneId(zoneName);
-        if (!zoneIdRet.ok()) {
-            LOG(ERROR) << "Zone " << zoneName << " not found";
-            handleErrorCode(cpp2::ErrorCode::E_NOT_FOUND);
-            onFinished();
-            return pickedHosts;
+StatusOr<std::unordered_map<HostAddr, int32_t>>
+CreateSpaceProcessor::getHostLoading() {
+    std::unique_ptr<kvstore::KVIterator> iter;
+    const auto& prefix = MetaServiceUtils::partPrefix();
```
bright-starry-sky, line 206: do we need to get all parts of space?

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -174,34 +200,97 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
+        return Status::Error("List Parts Failed");
+    }
+
+    std::unordered_map<HostAddr, int32_t> result;
+    while (iter->valid()) {
+        auto hosts = MetaServiceUtils::parsePartVal(iter->val());
+        for (auto& host : hosts) {
+            result[host]++;
         }
+        iter->next();
```
bright-starry-sky, line 219: Actually, I don't really understand this logic, means get all part count of each host?

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
darionyaphet: I will calculation the loading of each host

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
darionyaphet: sort by value

src/meta/processors/zoneMan/AddZoneProcessor.cpp
```diff
@@ -14,20 +16,31 @@ void AddZoneProcessor::process(const cpp2::AddZoneReq& req) {
...
+    for (auto& node : nodes) {
```
liuyu85cn, line 34: may be use std::includes ?

darionyaphet: Good catch & fixed

liuyu85cn (dismissed):

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -174,34 +200,97 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
+}
 
+StatusOr<std::vector<HostAddr>>
+CreateSpaceProcessor::pickHostsWithZone(const std::vector<std::string>& zones,
+                                        std::unordered_map<HostAddr, int32_t>& loading) {
+    std::vector<HostAddr> pickedHosts;
+    for (auto& zoneName : zones) {
         auto zoneKey = MetaServiceUtils::zoneKey(zoneName);
         auto zoneValueRet = doGet(std::move(zoneKey));
+        auto hosts = MetaServiceUtils::parseZoneHosts(std::move(zoneValueRet).value());
```
critical27, line 231: No need to read so many times.

darionyaphet: Do you think it's really necessary? Twice I can take it.

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -162,46 +190,113 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
...
+        auto zoneKey = MetaServiceUtils::zoneKey(zone);
         auto zoneValueRet = doGet(std::move(zoneKey));
         if (!zoneValueRet.ok()) {
-            LOG(ERROR) << "Get zone " << zoneName << " failed";
-            handleErrorCode(cpp2::ErrorCode::E_STORE_FAILURE);
-            onFinished();
-            return pickedHosts;
+            LOG(ERROR) << "Get zone " << zone << " failed";
+            return Status::Error("Get Zone Failed");
         }
```
critical27, line 272: These could move out from `pickLightLoadZones`. No need to read for each partition.

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -126,8 +126,60 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
...
+            auto hosts = MetaServiceUtils::parseZoneHosts(std::move(zoneValueRet).value());
+            for (auto& host : hosts) {
+                auto hostIter = hostLoading.find(host);
+                if (hostIter == hostLoading.end()) {
+                    LOG(ERROR) << "Host " << host << " not found";
+                    handleErrorCode(cpp2::ErrorCode::E_NOT_FOUND);
+                    onFinished();
+                    return;
+                }
+                zoneLoading[zone] += hostIter->second;
```
critical27, line 159: emm, if you move this part as well, `zoneLoading` won't be accurate anymore.

critical27 (approved):
LGTM

liuyu85cn (approved):

PR #188, opened 2020-10-29
Support space level statistics

panda-sheep:
Depend on common pr #286

### DIFF
--- a/src/meta/MetaServiceUtils.cpp
+++ b/src/meta/MetaServiceUtils.cpp
@@ -1068,7 +1068,7 @@ std::string MetaServiceUtils::statisKey(GraphSpaceID spaceId) {
     return key;
 }
 
-std::string MetaServiceUtils::statisVal(const cpp2::StatisItem &statisItem) {
+std::string MetaServiceUtils::statisVal(const cpp2::StatisItem& statisItem) {
     std::string val;
     apache::thrift::CompactSerializer::serialize(statisItem, &val);
     return val;
--- a/src/meta/MetaServiceUtils.cpp
+++ b/src/meta/MetaServiceUtils.cpp
@@ -30,7 +30,10 @@ const std::string kLeadersTable        = "__leaders__";          // NOLINT
 const std::string kGroupsTable         = "__groups__";           // NOLINT
 const std::string kZonesTable          = "__zones__";            // NOLINT
 const std::string kListenerTable       = "__listener__";         // NOLINT
-
+// Used to record the number of vertices and edges in the space
+// The number of vertices of each tag in the space
+// The number of edges of each edgetype in the space
+const std::string kStatisTable         = "__statis__";           // NOLINT
 
 const std::string kHostOnline  = "Online";       // NOLINT
 const std::string kHostOffline = "Offline";      // NOLINT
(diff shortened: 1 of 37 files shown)


src/storage/admin/StatisTask.cpp
```diff
@@ -0,0 +1,230 @@
...
+                spaceEdges++;
+                edgetypeEdges[edgeType] += 1;
+                lastSrcVertexId = source;
+                lastEdgeType  = edgeType;
+                lastRank = ranking;
+                lastDstVertexId = destination;
+            }
+        }
+        iter->next();
+        continue;
```
liuyu85cn, line 174: unnecesssry continue

src/storage/admin/StatisTask.cpp
```diff
@@ -0,0 +1,230 @@
...
+                } else {
+                    tagsVertices[tagId] += 1;
+                    lastTagId = tagId;
+                }
+            } else {
+                tagsVertices[tagId] += 1;
+                spaceVertices++;
+                lastTagId = tagId;
+                lastVertexId  = vId;
+            }
```
bright-starry-sky, line 139: A question，Is remove duplicate vid for `spaceVertices` at here? for example : 
tag1 (c1)
tag2 (c1)

vertex1 : tag1(c1) ---> count(v) of tag1 == 1
vertex1: tag2(c1) ---> count(v) of tag2 == 1
so the spaceVertices == 1 ? or 2?

src/storage/admin/StatisTask.cpp
panda-sheep: Good！

src/storage/admin/StatisTask.cpp
panda-sheep: Good question.
First of all, `spaceVertices` is 1 as in your case above.

Here use `lastVertexId` and current vid to remove duplicate data.
You can refer to the case on line 105.

bright-starry-sky: > Good question.
> First of all, `spaceVertices` is 1 as in your case above.
> 
> Here use `lastVertexId` and current vid to remove duplicate data.
> You can refer to the case on line 105.

got, nice.

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -387,19 +394,24 @@ void AdminClient::getResponse(std::vector<HostAddr> hosts,
...
                 p.setValue(Status::Error(folly::stringPrintf("RPC failure in AdminClient: %s",
                                                              t.exception().what().c_str())));
                 return;
             }
-            auto&& result = std::move(t).value().get_result();
+            auto&& result = std::move(t.value().get_result());
             if (result.get_failed_parts().empty()) {
+                // succeeded
                 p.setValue(Status::OK());
+                if (statisResult && t.value().__isset.statis) {
```
critical27, line 412: This is so weird - -. Maybe use template, or at least just add a new function instead of effect others.

src/meta/MetaServiceUtils.cpp
```diff
@@ -1057,5 +1060,29 @@ PartitionID MetaServiceUtils::parseListenerPart(folly::StringPiece rawData) {
     return *reinterpret_cast<const PartitionID*>(rawData.data() + offset);
 }
 
+std::string MetaServiceUtils::statisKey(GraphSpaceID spaceId) {
```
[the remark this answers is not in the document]
panda-sheep: Now only the result of the last execution can be read.
I think the historical results are meaningless.

src/meta/processors/admin/AdminClient.cpp
panda-sheep: The `Status `is currently returned. When all `Status `are ok, start statistics
`cpp2::StatisItem.`

What you said makes sense, but in that case, more code changes.
I think the smallest code changed like this. WDYT?

src/storage/admin/StatisTask.cpp
```diff
@@ -0,0 +1,229 @@
...
+        return kvstore::ResultCode::ERR_SPACE_NOT_FOUND;
+    }
+    auto vIdLen = vIdLenRet.value();
+
+    LOG(INFO) << "Start statis task";
+    CHECK_NOTNULL(env_->kvstore_);
+    auto prefix = NebulaKeyUtils::partPrefix(part);
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto ret = env_->kvstore_->prefix(spaceId, part, prefix, &iter);
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
```
critical27, line 76: You need to bypass the leader_change, even it is not leader, still read it.

src/meta/processors/admin/AdminClient.cpp
critical27: At least don't add a `cpp2::StatisItem* statisResult`, you could pass a function to do whatever you like. Other methods in `AdminClient` just pass an empty function. If you just add a pointer, subsequent changes will add more and more pointer.

src/storage/admin/StatisTask.cpp
panda-sheep: Got,  I will open a new pr to handle it.

panda-sheep: Fixed, thanks for your reminder！

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -363,26 +374,29 @@ folly::Future<Status> AdminClient::getResponse(
...
 
-template<typename Request, typename RemoteFunc>
+template<typename Request,
+         typename RemoteFunc,
+         typename RespGenerator>
 void AdminClient::getResponse(std::vector<HostAddr> hosts,
                               int32_t index,
                               Request req,
                               RemoteFunc remoteFunc,
+                              RespGenerator respGen,
```
liuyu85cn, line 384: better use folly::Optional?

panda-sheep: Good point，👍.

critical27 (approved):
LGTM, job deduplication will open another PR

liuyu85cn (approved):
Good job, LGTM

critical27 (approved):

PR #190, opened 2020-10-29
Backup/restore implementation

monadbobo:
Functions currently implemented:

1. added an API for full backup (createBackup).
2. reformat the partial code.
3. modify snapshot interface.

depend https://github.com/vesoft-inc/nebula-common/pull/344


src/meta/test/CreateBackupProcessorTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "common/fs/TempDir.h"
+#include "meta/processors/admin/CreateBackupProcessor.h"
+#include "meta/test/TestUtils.h"
+#include "utils/Utils.h"
+
+namespace nebula {
+namespace meta {
+
+#define RETURN_OK(req)                                                                             \
+    UNUSED(req);                                                                                   \
```
bright-starry-sky, line 18: Why is the parameter specified but not used?

src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -0,0 +1,286 @@
...
+    files.insert(files.end(),
+                 std::make_move_iterator(value(ret).begin()),
+                 std::make_move_iterator(value(ret).end()));
+
+    return files;
+}
+
+folly::Optional<std::unordered_set<GraphSpaceID>> CreateBackupProcessor::spaceNameToId(
+    const std::vector<std::string>* backupSpaces) {
+    std::unordered_set<GraphSpaceID> spaces;
```
bright-starry-sky, line 83: Now ， is blocking date write for meta？

src/kvstore/RocksEngine.cpp
```diff
@@ -461,5 +433,69 @@ ResultCode RocksEngine::createCheckpoint(const std::string& name) {
     return ResultCode::SUCCEEDED;
 }
 
-}  // namespace kvstore
-}  // namespace nebula
+ErrorOr<ResultCode, std::string> RocksEngine::backupTable(
+    const std::string& name,
+    const std::string& tablePrefix,
+    std::function<bool(const folly::StringPiece& key)> filter) {
+    auto backupPath = folly::stringPrintf(
```
critical27, line 440: How to tell the difference between multiple backup? BTW, why we don't write one sst for on space instead of multiple

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -600,24 +600,43 @@ folly::Future<Status> AdminClient::getLeaderDist(HostLeaderMap* result) {
...
     auto f = pro.getFuture();
-    getResponse({Utils::getAdminAddrFromStoreAddr(host)}, 0, std::move(req),
-                [] (auto client, auto request) {
-                    return client->future_createCheckpoint(request);
-                },
-                0, std::move(pro), 3 /*The snapshot operation need to retry 3 times*/);
+
+    auto* evb = ioThreadPool_->getEventBase();
+    auto storageHost = Utils::getAdminAddrFromStoreAddr(host);
+    folly::via(evb, [evb, storageHost, pro = std::move(pro), spaceId, name, this]() mutable {
```
critical27, line 615: Why don't use `getResponse`?

src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    if (!result.value()) {
+        handleErrorCode(cpp2::ErrorCode::E_BACKUP_BUILDING_INDEX);
+        onFinished();
+        return;
+    }
+
+    folly::SharedMutex::WriteHolder wHolder(LockUtils::snapshotLock());
+
+    auto hosts = ActiveHostsMan::getActiveHosts(kvstore_);
```
critical27, line 157: Do you think we need to check all hosts online before we start? We didn't do it in 1.0

src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -0,0 +1,277 @@
...
+    std::vector<kvstore::KV> data;
+    auto backupName = folly::format("BACKUP_{}", MetaServiceUtils::genTimestampStr()).str();
+
+    data.emplace_back(MetaServiceUtils::snapshotKey(backupName),
+                      MetaServiceUtils::snapshotVal(cpp2::SnapshotStatus::INVALID,
+                                                    NetworkUtils::toHostsStr(hosts)));
+    Snapshot::instance(kvstore_, client_)->setSpaces(spaces);
+
+    // step 1 : Blocking all writes action for storage engines.
+    auto ret = Snapshot::instance(kvstore_, client_)->blockingWrites(SignType::BLOCK_ON);
```
critical27, line 183: We don't need to cancel block write when we fail at each step?

src/meta/processors/admin/CreateSnapshotProcessor.cpp
```diff
@@ -69,9 +69,9 @@ void CreateSnapshotProcessor::process(const cpp2::CreateSnapshotReq&) {
 
     // step 3 : Create checkpoint for all storage engines and meta engine.
     auto csRet = Snapshot::instance(kvstore_,  client_)->createSnapshot(snapshot);
-    if (csRet != cpp2::ErrorCode::SUCCEEDED) {
+    if (csRet.isLeftType()) {
```
critical27, line 72: The old snapshot and backup file could be removed?

util/br/pkg/backup/backup.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		return err
+	}
+	binaryOut.Flush()
+	return nil
+}
+
+func (b *Backup) BackupCluster() error {
+	b.client = metaclient.NewMetaClient(b.log)
+	err := b.client.Open(b.config.MetaAddrs[0])
```
critical27, line 139: We could `use space` before backup, so we can try to get the leader.

util/br/pkg/restore/restore.go
```diff
@@ -0,0 +1,373 @@
...
+
+func (r *Restore) RestoreCluster() error {
+
+	err := r.downloadMetaFile()
+	if err != nil {
+		r.log.Error("download meta file failed", zap.Error(err))
+		return err
+	}
+
+	m, err := r.restoreMetaFile()
```
critical27, line 300: What is in the meta file for now?

util/br/pkg/restore/restore.go
```diff
@@ -0,0 +1,373 @@
...
+func stringToHostAddr(host string) (*nebula.HostAddr, error) {
+	ipAddr := strings.Split(host, ":")
+	port, err := strconv.ParseInt(ipAddr[1], 10, 32)
+	if err != nil {
+		return nil, err
+	}
+	return &nebula.HostAddr{ipAddr[0], nebula.Port(port)}, nil
+}
+
+func (r *Restore) restoreMeta(sstFiles []string, storageIDMap map[string]string, count int) error {
```
critical27, line 156: Why we can only restore only one meta, or I misunderstand it?

util/br/pkg/restore/restore.go
```diff
@@ -0,0 +1,373 @@
...
+		return err
+	}
+
+	err = r.stopCluster()
+	if err != nil {
+		r.log.Error("stop cluster failed", zap.Error(err))
+		return err
+	}
+
+	err = r.cleanupOriginal()
```
critical27, line 319: Better to move to the final step

src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -0,0 +1,277 @@
...
+//                                                             const std::vector<GraphSpaceID>&,
+//                                                             const std::string&)>
+static const std::unordered_map<std::string,
+                                std::function<decltype(MetaServiceUtils::backupSpaceTable)>>
+    tableMaps = {{"spaces", MetaServiceUtils::backupSpaceTable},
+                 {"parts", MetaServiceUtils::backupPartsTable},
+                 {"tags", MetaServiceUtils::backupTagsTable},
+                 {"edges", MetaServiceUtils::backupEdgesTable},
+                 {"indexes", MetaServiceUtils::backupIndexesTable},
+//                 {"index", MetaServiceUtils::backupIndexTable},
```
liuyu85cn, line 24: remove this ?

src/meta/MetaServiceUtils.cpp
```diff
@@ -951,6 +961,233 @@ HostAddr MetaServiceUtils::deserializeHostAddr(folly::StringPiece raw) {
...
+            if (type == EntryType::SPACE) {
+                auto sn = key.subpiece(kIndexTable.size() + sizeof(EntryType),
+                                       key.size() - kIndexTable.size() - sizeof(EntryType))
+                              .str();
+                LOG(INFO) << "sn was " << sn;
+                auto it = std::find_if(spaceName->cbegin(), spaceName->cend(), [&sn](auto& name) {
+                    if (sn == name) {
+                        return true;
+                    }
+                    return false;
```
bright-starry-sky, line 1076: return sn == name;

src/meta/processors/admin/RestoreProcessor.cpp
```diff
@@ -0,0 +1,51 @@
...
+                onFinished();
+                return;
+            }
+        }
+    }
+
+    for (auto &f : files) {
+        unlink(f.c_str());
+    }
+
```
panda-sheep, line 47: delete the backed up meta data after meta restore?
I feel that this is the user's own operation

src/meta/test/CreateBackupProcessorTest.cpp
monadbobo: Because there is no need to parse req here for testing purposes.

src/meta/processors/admin/CreateBackupProcessor.cpp
monadbobo: Yes, it need to use spacelock here.

src/meta/processors/admin/RestoreProcessor.cpp
monadbobo: At this point it means that the restore was completely successful, so I think it is possible to delete the temporary files.

src/kvstore/NebulaStore.cpp
```diff
@@ -889,15 +894,15 @@ ResultCode NebulaStore::createCheckpoint(GraphSpaceID spaceId, const std::string
...
+            auto walPath = folly::stringPrintf(
+                "%s/checkpoints/%s/wal/%d", engine->getDataRoot(), name.c_str(), part);
             auto p = nebula::value(ret);
             if (!p->linkCurrentWAL(walPath.data())) {
                 return ResultCode::ERR_CHECKPOINT_ERROR;
             }
         }
     }
-    return ResultCode::SUCCEEDED;
+    return cpPath;
```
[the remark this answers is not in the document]
monadbobo: This is because during backup, we need to get the path of each machine's temporary file (checkpoint) in order to upload it to the remote (s3/oss...).

src/meta/processors/admin/CreateSnapshotProcessor.cpp
monadbobo: Yes, I am now calling the dropsnapshot interface to clean up temporary files when backup is done.

src/meta/MetaServiceUtils.cpp
```diff
@@ -951,6 +961,233 @@ HostAddr MetaServiceUtils::deserializeHostAddr(folly::StringPiece raw) {
...
+    folly::SharedMutex::ReadHolder rHolder(LockUtils::spaceLock());
+    auto prefix = rebuildIndexStatusPrefix();
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto ret = kvstore->prefix(kDefaultSpaceId, kDefaultPartId, prefix, &iter);
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
+        return folly::none;
+    }
+
+    while (iter->valid()) {
+        if (iter->val() == "RUNNING") {
```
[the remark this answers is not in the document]
monadbobo: Do you mean that it need to type an error log here?

src/kvstore/NebulaStore.cpp
```diff
@@ -955,7 +960,7 @@ ResultCode NebulaStore::setWriteBlocking(GraphSpaceID spaceId, bool sign) {
             }
             if (ret != ResultCode::SUCCEEDED) {
                  LOG(ERROR) << "Part sync failed. space : " << spaceId << " Part : " << part;
-                 return ret;
+                 continue;
```
bright-starry-sky, line 963: Why ignore sync errors?

src/meta/processors/admin/AdminClient.cpp
monadbobo: This is because the current createSnapshot interface needs to return the corresponding snapshot path.

util/br/pkg/restore/restore.go
monadbobo: You can see the contents of the corresponding meta file in the [PR of nebula-common](https://github.com/vesoft-inc/nebula-common/pull/288).

util/br/pkg/restore/restore.go
monadbobo: We have to leave it here for now, because we are downloading the meta and storage backups directly to their data directories, and after this step, we will download the content directly from the remote.

src/kvstore/NebulaStore.cpp
```diff
@@ -1051,6 +1056,58 @@ void NebulaStore::cleanWAL() {
...
+ResultCode NebulaStore::restoreFromFiles(GraphSpaceID spaceId,
+                                         const std::vector<std::string>& files) {
+    auto spaceRet = space(spaceId);
+    if (!ok(spaceRet)) {
+        LOG(ERROR) << "Get Space " << spaceId << " Failed";
+        return error(spaceRet);
+    }
+    auto space = nebula::value(spaceRet);
+
+    for (auto& engine : space->engines_) {
```
critical27, line 1101: `restoreFromFiles` is both called by storage and meta? If storage call this method, it could be wrong if multi data_path is specified.

src/kvstore/RocksEngine.cpp
monadbobo: Just because the order of the keys must be guaranteed when exporting sst, and the space id in some of our tables is not a prefix, it is easier to export sst by table.

src/kvstore/NebulaStore.cpp
monadbobo: "restoreFromFiles" will only be called by Meta.

src/kvstore/NebulaStore.cpp
monadbobo: The error is ignored here because the meta call to the setWriteBlocking interface actually ignores the error.

src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -0,0 +1,277 @@
...
+                 {"index_status", MetaServiceUtils::backupIndexStatusTable},
+                 {"users", MetaServiceUtils::backupUsersTable},
+                 {"roles", MetaServiceUtils::backupRolesTable},
+                 {"configs", MetaServiceUtils::backupConfigsTable}};
```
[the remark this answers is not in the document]
monadbobo: Yes, but don't want to change too much code at the moment (need to refactor MetaServiceUtils code).

src/meta/processors/admin/CreateSnapshotProcessor.cpp
critical27: I mean we could just remove the file `CreateSnapshotProcessor.cpp`? The `br` tool only call create backup?

bright-starry-sky:
A question，If just backup one space, checkpoint all spaces in the storage layer and then copy one space , right?

src/kvstore/NebulaStore.cpp
monadbobo: > I don't quite understand, if sync return error, it can't be ignored.

This is because in setWriteBlocking, it goes through all the parts and then sets the block flag, and finally sends a sync, where the sync must have a part of the failure (because the leader change), and I think this part of the failure can be ignored, because the failure of the machine will eventually call setWriteBlocking in the meta.

src/meta/processors/admin/CreateSnapshotProcessor.cpp
monadbobo: > I mean we could just remove the file `CreateSnapshotProcessor.cpp`? The `br` tool only call create backup?

I think it's okay to keep it for now, but we can remove CreateSnapshotProcessor.cpp when the BR is fully stabilized.

monadbobo:
> A question，If just backup one space, checkpoint all spaces in the storage layer and then copy one space , right?

yes.

src/meta/processors/admin/SnapShot.cpp
```diff
@@ -63,9 +67,13 @@ cpp2::ErrorCode Snapshot::blockingWrites(storage::cpp2::EngineSignType sign) {
     auto spacesHosts = retSpacesHosts.value();
     for (const auto& spaceHosts : spacesHosts) {
         for (const auto& host : spaceHosts.second) {
+            LOG(INFO) << "will block write host: " << host;
             auto status = client_->blockingWrites(spaceHosts.first, sign, host).get();
             if (!status.ok()) {
                 LOG(ERROR) << " Send blocking sign error on host : " << host;
+                if (sign == storage::cpp2::EngineSignType::BLOCK_ON) {
```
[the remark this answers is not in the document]
monadbobo: This is because when deleting the write flag, we need to delete as many flags as possible.

src/kvstore/RocksEngine.cpp
monadbobo: Yes, it may need to refactor the MetaServiceUtils code about table here.

monadbobo: Or we need to redesign the table structure so that one type of table has nothing to do with spaceid and one type of table can get all the data directly from the spaceid prefix.

src/meta/processors/admin/AdminClient.cpp
critical27: `getResponse` could support it, there is a `respGen`.

src/meta/processors/admin/SnapShot.cpp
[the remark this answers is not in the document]
critical27: You said that 
> This is because in setWriteBlocking, it goes through all the parts and then sets the block flag, and finally sends a sync, where the sync must have a part of the failure (because the leader change), and I think this part of the failure can be ignored, because the failure of the machine will eventually call setWriteBlocking in the meta.

I believe this is where meta call setWriteBlocking, I don't think we can ignore failure in any case (If you ignore failure when unblock, the part can't service anymore). We'd better modify `NebulaStore::setWriteBlocking`.

src/kvstore/NebulaStore.cpp
critical27: I got it. I suppose we don't need to `sync` because this is a memory lock, and when meta `setWriteBlocking`, it send to all storage in a space. @bright-starry-sky , do you remember why we sync here?

src/meta/processors/admin/AdminClient.cpp
monadbobo: Neither function of getResponse satisfies the condition. The first function (4 arguments) requires that Request must contain a part id, and the second requires that the return value must be "Status".

src/meta/processors/admin/SnapShot.cpp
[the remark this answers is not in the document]
monadbobo: > You said that
> 
> > This is because in setWriteBlocking, it goes through all the parts and then sets the block flag, and finally sends a sync, where the sync must have a part of the failure (because the leader change), and I think this part of the failure can be ignored, because the failure of the machine will eventually call setWriteBlocking in the meta.
> 
> I believe this is where meta call setWriteBlocking, I don't think we can ignore failure in any case (If you ignore failure when unblock, the part can't service anymore). We'd better modify `NebulaStore::setWriteBlocking`.

Okay, here I think we need to call setWriteBlocking as much as possible, but we need to keep the error return to the upper level.

src/meta/processors/admin/CreateBackupProcessor.cpp
monadbobo: I think it is needed because if some machine is offline at the time of the backup, then it is possible that the backup will fail.

src/kvstore/NebulaStore.cpp
monadbobo: Yeah, I thought it was weird the first time I saw the code here, and when I tested it, sync must have returned an error because the current machine is definitely not the leader of all the parts.

src/kvstore/RocksEngine.cpp
```diff
@@ -461,5 +433,69 @@ ResultCode RocksEngine::createCheckpoint(const std::string& name) {
...
+    s = sstFileWriter.Finish();
+    if (!s.ok()) {
+        LOG(WARNING) << "Failure to insert data when backupTable,  " << backupPath
+                     << ", error: " << s.ToString();
+        return ResultCode::ERR_BACKUP_EMPTY_TABLE;
+    }
+
+    if (sstFileWriter.FileSize() == 0) {
+        return ResultCode::ERR_BACKUP_EMPTY_TABLE;
+    }
```
bright-starry-sky, line 495: Are empty tables allowed? for example roles table.

src/meta/processors/admin/CreateBackupProcessor.cpp
critical27: If so, this need to changed. `ActiveHostsMan::getActiveHosts` only get all active hosts, not the hosts contain specified spaces.

src/kvstore/NebulaStore.cpp
critical27: If so we could remove it, all failure should return to meta.

src/meta/MetaServiceUtils.cpp
```diff
@@ -955,6 +1021,249 @@ HostAddr MetaServiceUtils::deserializeHostAddr(folly::StringPiece raw) {
...
+    const std::string& backupName,
+    const std::vector<std::string>* spaceName) {
+    return kvstore->backupTable(
+        kDefaultSpaceId,
+        backupName,
+        kIndexTable,
+        [spaces, spaceName](const folly::StringPiece& key) -> bool {
+            if (spaces.empty() || spaceName == nullptr) {
+                return false;
+            }
```
bright-starry-sky, line 1083: || spaceName->empty() ?

src/kvstore/RocksEngine.cpp
monadbobo: Here the sstfilewriter is empty which means that the table does not exist.

src/meta/MetaServiceUtils.cpp
monadbobo: yes, I will fix this.

critical27 (approved):
Generally LGTM

panda-sheep (commented):
Good job.

src/kvstore/NebulaStore.cpp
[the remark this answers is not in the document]
panda-sheep: BTW,  here we add a TODO to the case of multiple datapaths

src/meta/processors/admin/AdminClient.cpp
panda-sheep: Got. thx

panda-sheep (approved):
Excellent！

PR #193, opened 2020-11-03
Handle path build expr.

CPWstatic:


critical27 (approved):

panda-sheep (approved):

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -488,6 +488,7 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkExp(const Expression* exp,
         case Expression::Kind::kUUID:
         case Expression::Kind::kVar:
         case Expression::Kind::kVersionedVar:
+        case Expression::Kind::kPathBuild:
         default: {
```
panda-sheep, line 492: Because there is no special logic to deal with this case, just use default, right?

CPWstatic: Yes

PR #199, opened 2020-11-10
Balance for zone

darionyaphet:


liuyu85cn:
how about add some description

darionyaphet:
> how about add some description

emmm.. It will be a long story

src/meta/processors/admin/Balancer.cpp
```diff
@@ -580,79 +693,85 @@ int32_t Balancer::acquireLeaders(
...
+        VLOG(3) << "Try acquire leader for part " << partId;
+        auto minLoad = partBounds_[partId].first;
         // find the leader of partId
         auto peers = peersMap[partId];
         for (const auto& peer : peers) {
             if (peer == host || !activeHosts.count(peer)) {
                 continue;
             }
             // if peer is the leader of partId and can transfer, then transfer it to host
+            // if the space specified with zone, we should check for conflict.
```
critical27, line 705: Where do we check conflict? Is this right?

src/meta/processors/admin/Balancer.cpp
```diff
@@ -580,79 +693,85 @@ int32_t Balancer::acquireLeaders(
...
+        // leader should move to the peer with lowest loading
+        auto min = std::min_element(peers.begin(), peers.end(),
+                                    [&](const auto &l, const auto &r) -> bool {
+                                        if (host == l || !activeHosts.count(l)) {
+                                            return false;
+                                        }
+                                        return leaderParts[l].size() < leaderParts[r].size();
+                                    });
+        // If peer can accept this partition leader, than host will transfer to the peer
+        // if the space specified with zone, we should check for conflict.
```
critical27, line 757: ditto,  delete the comment

src/meta/processors/admin/Balancer.cpp
```diff
@@ -675,5 +794,84 @@ void Balancer::simplifyLeaderBalnacePlan(GraphSpaceID spaceId, LeaderBalancePlan
...
+    if (code != kvstore::ResultCode::SUCCEEDED) {
+        LOG(ERROR) << "List Parts Failed";
+        return false;
+    }
+
+    std::unordered_map<HostAddr, int32_t> hostLoading;
+    while (iter->valid()) {
+        auto hosts = MetaServiceUtils::parsePartVal(iter->val());
+        for (auto& host : hosts) {
+            hostLoading[host]++;
```
critical27, line 810: I don't think we should count for all hosts, only hosts in specified group should be counted.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -675,5 +794,84 @@ void Balancer::simplifyLeaderBalnacePlan(GraphSpaceID spaceId, LeaderBalancePlan
...
+        }
+        iter->next();
+    }
+
+    for (auto it = hostLoading.begin(); it != hostLoading.end(); it++) {
+        int32_t min = it->second / replicaFactor;
+        auto pair = Bounds(min, min + 1);
```
critical27, line 817: Would you explain more about it?

critical27 (commented):
Please add some description about algorithm of leader balance in summary, especially how the bound is calculated when space is created by zone.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -675,5 +794,84 @@ void Balancer::simplifyLeaderBalnacePlan(GraphSpaceID spaceId, LeaderBalancePlan
...
+            auto hostIter = hostLoading.find(address);
+            if (hostIter == hostLoading.end()) {
+                LOG(ERROR) << "Can't find host " << address;
+                return false;
+            }
+
+            VLOG(3) << "Part: " << part << " size: " << hostIter->second;
+            partLoading[part].emplace_back(hostIter->second);
+        }
+    }
```
critical27, line 858: Don't do disk io and calculate for each time one partition is transferred.

src/meta/MetaServiceUtils.cpp
```diff
@@ -35,6 +35,8 @@ const std::string kListenerTable       = "__listener__";         // NOLINT
 // The number of vertices of each tag in the space
 // The number of edges of each edgetype in the space
 const std::string kStatisTable         = "__statis__";           // NOLINT
+const std::string kBalanceTaskTable    = "__balance_task__";     // NOLINT
+const std::string kBalancePlanTable    = "__balance_plan__";     // NOLINT
```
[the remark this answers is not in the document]
darionyaphet: maybe need to be discussed later

src/meta/processors/admin/Balancer.cpp
```diff
@@ -675,5 +794,84 @@ void Balancer::simplifyLeaderBalnacePlan(GraphSpaceID spaceId, LeaderBalancePlan
...
+        auto sum = std::accumulate(partIter->second.begin(), partIter->second.end(), 0);
+        int32_t min = sum / size / replicaFactor;
+        int32_t max = min;
+        if (sum % (size * replicaFactor) != 0) {
+            max += 1;
+        }
+
+        auto pair = Bounds(min, max);
+        LOG(INFO) << "Part: " << part << " Bounds: [" << pair.first << ", " << pair.second << "]";
+        partBounds_[part] = std::move(pair);
```
critical27, line 872: Do you think the bound is too aggressive? Maybe a bigger range will be more safe. For example, given a part, three hosts hold it, I think the a range that covers average of the given three hosts looks more reasonable. `min < avg < max`. 

I am not pretty sure about that we should get it by `sum / size / replicaFactor`. A extreme example:
A: part [1]
B: part [1]
C: part [1-100]
They both have part 1, but we can't make promise that for the hosts which have part 1 should be in the range `sum / size / replicaFactor`, in this example, that would be `102 / 3 /3`.

src/meta/processors/admin/Balancer.cpp
darionyaphet: this message has discard

src/meta/processors/admin/Balancer.cpp
```diff
@@ -345,23 +358,60 @@ void Balancer::getHostParts(GraphSpaceID spaceId,
...
+            totalHosts.insert(totalHosts.end(), hosts.begin(), hosts.end());
+        }
+
+        for (auto hostIter = hostParts.begin(); hostIter != hostParts.end(); hostIter++) {
+            auto found = std::find(totalHosts.begin(), totalHosts.end(), hostIter->first);
+            if (found == totalHosts.end()) {
+                delete &hostIter->second;
+                hostParts.erase(hostIter);
+            }
+        }
```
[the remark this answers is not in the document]
darionyaphet: yeah it's not a good idea so I have supported a method in `ActiveHostManager` which is named `getActiveHostsInGroup`

src/meta/processors/admin/Balancer.cpp
darionyaphet: Not all hosts, it's only the space-related hosts

src/meta/processors/admin/Balancer.cpp
darionyaphet: I‘m not sure `[0.95 * avg, 1.05 * avg]` is a bigger range.  Actually they are the same ([min, max])

src/meta/processors/admin/Balancer.cpp
darionyaphet: really good idea

src/meta/processors/admin/Balancer.cpp
```diff
@@ -471,18 +509,27 @@ cpp2::ErrorCode Balancer::leaderBalance() {
                 }
             }
         }).wait();
-        LOG(INFO) << failed << " partiton failed to transfer leader";
+
         inLeaderBalance_ = false;
-        return cpp2::ErrorCode::SUCCEEDED;
+        if (failed != 0) {
```
critical27, line 514: just return SUCCEEDED, we don't expect all part transferred ok.

critical27 (commented):
Generally LGTM

src/meta/processors/admin/Balancer.cpp
```diff
@@ -580,79 +671,85 @@ int32_t Balancer::acquireLeaders(
...
+                                        if (host == l || !activeHosts.count(l)) {
+                                            return false;
+                                        }
+                                        return leaderParts[l].size() < leaderParts[r].size();
+                                    });
+        // If peer can accept this partition leader, than host will transfer to the peer
+        // if the space specified with zone, we should check for conflict.
+        auto& peerLeaders = leaderParts[*min];
+        int32_t leaderCount = peerLeaders.size();
+        if (leaderCount < hostBounds_[*min].second) {
```
critical27, line 738: Why compare with `minLoad`, aka `partBounds_[partId].first` in line 689, and in this line was compared with `hostBounds_[*min].second`? I suppose we only one of `partBounds_` and `hostBounds_`

src/meta/ActiveHostsMan.cpp
```diff
@@ -91,6 +91,38 @@ std::vector<HostAddr> ActiveHostsMan::getActiveHostsInZone(kvstore::KVStore* kv,
...
+        return activeHosts;
+    }
+
+    auto zoneNames = MetaServiceUtils::parseZoneNames(std::move(groupValue));
+    for (const auto& zoneName : zoneNames) {
+        auto hosts = getActiveHostsInZone(kv, zoneName, expiredTTL);
+        LOG(INFO) << "Zone " << zoneName << " Host Size: " << hosts.size();
+        activeHosts.insert(activeHosts.end(), hosts.begin(), hosts.end());
+    }
+    return activeHosts;
```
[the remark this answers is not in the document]
darionyaphet: Only when space build on group needs to filter out some host when not included in the group. when it be calling, it must have group name in the space property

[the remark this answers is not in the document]
critical27: Yeah, but the function name did not imply that, maybe change it will be better.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -471,188 +507,229 @@ cpp2::ErrorCode Balancer::leaderBalance() {
...
-                    plan.emplace_back(spaceId, partId, peer, host);
-                    LOG(INFO) << "plan trans leader: " << spaceId << " " << partId << " from "
-                              << peer.host
-                              << peer.port << " to "
-                              << host.host
-                              << ":" << host.port;
-                    ++taskCount;
-                    break;
-                }
+            // if the space specified with zone, we should check for conflict.
```
critical27, line 665: This comment could be removed

src/meta/processors/admin/Balancer.cpp
```diff
@@ -471,188 +507,229 @@ cpp2::ErrorCode Balancer::leaderBalance() {
...
+        // leader should move to the peer with lowest loading
+        auto target = std::min_element(targets.begin(), targets.end(),
+                                       [&](const auto &l, const auto &r) -> bool {
+                                           if (source == l || !activeHosts.count(l)) {
+                                               return false;
+                                           }
+                                           return leaderParts[l].size() < leaderParts[r].size();
+                                       });
+        // If peer can accept this partition leader, than host will transfer to the peer
+        // if the space specified with zone, we should check for conflict.
```
critical27, line 716: ditto

critical27 (commented):
Good job

src/meta/processors/admin/Balancer.cpp
darionyaphet: fixed

critical27 (approved):
Good job!

liuyu85cn (approved):
LGTM

PR #158, opened 2020-10-10
Support string type index to add | delete | update | scan

bright-starry-sky:
depend on #156

PR #200, opened 2020-11-10
Fix crash for string property index which has been created before new index syntax

yixinglu:
When use string column index created before #158 , nebula storage will crash since dereference a nullptr on column type length.


critical27 (approved):
Good catch, btw, graph need to catch it as well, especially old index before #158, old string index don't have fixed size.

bright-starry-sky (approved):
In fact, all indexes should be re-create and re-build; the data structure of this index has changed since string type supported. better note in doc.
This  error will not appear after recreate.
Good fix. LGTM.

PR #202, opened 2020-11-11
fix job manager may remove queue jobs

liuyu85cn:
if a job is in queue, 
job manager will not set its start time. 
and this job may removed at show jobs


panda-sheep (approved):
Well done!

jievince (approved):

PR #203, opened 2020-11-11
Fix balance failed

laura-ding:
https://discuss.nebula-graph.com.cn/t/topic/1449

### DIFF
--- a/src/meta/processors/admin/Balancer.cpp
+++ b/src/meta/processors/admin/Balancer.cpp
@@ -471,11 +471,8 @@ cpp2::ErrorCode Balancer::leaderBalance() {
                 }
             }
         }).wait();
+        LOG(INFO) << failed << " partiton failed to transfer leader";
         inLeaderBalance_ = false;
-        if (failed > 0) {
-            LOG(ERROR) << failed << " partiton failed to transfer leader";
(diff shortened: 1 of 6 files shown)


critical27 (dismissed):
Gooood catch, thanks taking care of it!

src/meta/processors/admin/Balancer.cpp
```diff
@@ -471,9 +471,12 @@ cpp2::ErrorCode Balancer::leaderBalance() {
                 }
             }
         }).wait();
-        LOG(ERROR) << failed << " partiton failed to transfer leader";
         inLeaderBalance_ = false;
-        return cpp2::ErrorCode::E_BALANCER_FAILURE;
+        if (failed > 0) {
```
panda-sheep, line 475: Good catch！

panda-sheep (approved):
Excellent！👏

src/meta/processors/admin/Balancer.cpp
critical27: Hold on, @laura-ding, this place is intended to do that, we don't expect all transfer leader is successful, no need to return failure.

critical27:
Hold on, don't merge plz

critical27 (approved):
Well done

bright-starry-sky (approved):
Well done.

PR #204, opened 2020-11-12
encode null as 0xFF instead of \0, support lookup filter out of index

critical27:
As title, add more test cases.

The problem of null as \0 is that look up always follow the rule that lookup in range [start, end). \0 will be scanned like `where tag.col < 10`. After change to 0xFF, null can't be scanned even if like query `where tag.col > 10`, because the range will be [10, max), which can't fix. Everything works fine except one, literally `0xFF .... 0xFF` (e.g. `INT_MAX`) is reserved. 

Null of string index length will be fixed in another PR.

### DIFF
--- a/src/storage/index/LookupBaseProcessor.inl
+++ b/src/storage/index/LookupBaseProcessor.inl
@@ -22,8 +22,20 @@ cpp2::ErrorCode LookupBaseProcessor<REQ, RESP>::requestCheck(const cpp2::LookupI
     planContext_->isEdge_ = indices.get_is_edge();
     if (planContext_->isEdge_) {
         planContext_->edgeType_ = indices.get_tag_or_edge_id();
+        auto edgeName = this->env_->schemaMan_->toEdgeName(spaceId_, planContext_->edgeType_);
+        if (!edgeName.ok()) {
+            return cpp2::ErrorCode::E_EDGE_NOT_FOUND;
+        }
+        planContext_->edgeName_ = std::move(edgeName.value());
+        schema_ = this->env_->schemaMan_->getEdgeSchema(spaceId_, planContext_->edgeType_);
     } else {
         planContext_->tagId_ = indices.get_tag_or_edge_id();
+        auto tagName = this->env_->schemaMan_->toTagName(spaceId_, planContext_->tagId_);
+        if (!tagName.ok()) {
+            return cpp2::ErrorCode::E_TAG_NOT_FOUND;
+        }
+        planContext_->tagName_ = std::move(tagName.value());
+        schema_ = this->env_->schemaMan_->getTagSchema(spaceId_, planContext_->tagId_);
     }
 
     if (indices.get_contexts().empty()) {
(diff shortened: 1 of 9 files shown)


panda-sheep (commented):
Can you give a simple example?

src/storage/index/LookupBaseProcessor.inl
```diff
@@ -22,8 +22,20 @@ cpp2::ErrorCode LookupBaseProcessor<REQ, RESP>::requestCheck(const cpp2::LookupI
...
+        planContext_->edgeName_ = std::move(edgeName.value());
+        schema_ = this->env_->schemaMan_->getEdgeSchema(spaceId_, planContext_->edgeType_);
     } else {
         planContext_->tagId_ = indices.get_tag_or_edge_id();
+        auto tagName = this->env_->schemaMan_->toTagName(spaceId_, planContext_->tagId_);
+        if (!tagName.ok()) {
+            return cpp2::ErrorCode::E_TAG_NOT_FOUND;
+        }
+        planContext_->tagName_ = std::move(tagName.value());
+        schema_ = this->env_->schemaMan_->getTagSchema(spaceId_, planContext_->tagId_);
```
panda-sheep, line 38: The judgment of if (schema_ == nullptr) should be here
 not in line 243

src/storage/index/LookupBaseProcessor.inl
```diff
@@ -316,21 +337,14 @@ LookupBaseProcessor<REQ, RESP>::buildPlanWithData(const cpp2::IndexQueryContext&
         return nullptr;
     }
 
-    auto schemaName = planContext_->isEdge_
-                      ? this->env_->schemaMan_->toEdgeName(spaceId_, planContext_->edgeType_)
-                      : this->env_->schemaMan_->toTagName(spaceId_, planContext_->tagId_);
-    if (!schemaName.ok()) {
-        return nullptr;
-    }
-
```
panda-sheep, line 325: line 312- line 315 Is not needed.
becase line 25 you have handled

src/storage/index/LookupBaseProcessor.inl
```diff
@@ -446,20 +460,14 @@ LookupBaseProcessor<REQ, RESP>::buildPlanWithDataAndFilter(const cpp2::IndexQuer
         return nullptr;
     }
```
panda-sheep, line 461: ditto

src/storage/index/LookupBaseProcessor.inl
critical27: That's because `schmea_` is not used in all cases, only when we need to read data.

src/storage/index/LookupBaseProcessor.inl
critical27: good catch

critical27 (commented):
> Can you give a simple example?

For example, only one int col nullable, if you encode null as 0x00, it will be `[0x00, 0x00, 0x00, 0x00], and it will be scanned with query `lookup ... where int_col < 10`

src/storage/index/LookupBaseProcessor.inl
panda-sheep: I understand what you mean, not all schama is used
Only `buildPlanWithData ` and `buildPlanWithDataAndFilter `   use `schema `to read data
But as you said, there are two small problems:
1) The schema is null or not, it needs to be judged in at least two places
2) `buildPlanBasic `and `buildPlanWithFilter ` only fetch and read data from the index, but when the schema does not exist, the index cannot exist

so the schema is null or not can be unified ?

panda-sheep (approved):
Excellent!

src/storage/index/LookupBaseProcessor.inl
critical27: Fixed

bright-starry-sky (approved):
Well done. Nice job.

PR #205, opened 2020-11-12
Add check wrong null type

laura-ding:


panda-sheep (approved):
Good！

panda-sheep (commented):
By the way, should add case to the integration test?

liuyu85cn (approved):

laura-ding:
> By the way, should add case to the integration test?

yeah, I will add in graph repo

PR #206, opened 2020-11-13
fix: task manager should continue not exit if some task gen sub taks …

liuyu85cn:
fix: task manager should continue not exit if some task gen sub taks …

### DIFF
--- a/src/storage/admin/AdminTaskManager.cpp
+++ b/src/storage/admin/AdminTaskManager.cpp
@@ -111,10 +111,11 @@ void AdminTaskManager::schedule() {
         auto task = it->second;
         auto errOrSubTasks = task->genSubTasks();
         if (!nebula::ok(errOrSubTasks)) {
-            LOG(ERROR) << "genSubTasks() failed=" << static_cast<int>(nebula::error(errOrSubTasks));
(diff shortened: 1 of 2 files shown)


src/storage/admin/AdminTaskManager.cpp
```diff
@@ -111,10 +111,11 @@ void AdminTaskManager::schedule() {
         auto task = it->second;
         auto errOrSubTasks = task->genSubTasks();
         if (!nebula::ok(errOrSubTasks)) {
-            LOG(ERROR) << "genSubTasks() failed=" << static_cast<int>(nebula::error(errOrSubTasks));
+            auto code = static_cast<int>(nebula::error(errOrSubTasks));
+            LOG(ERROR) << "genSubTasks() failed, err=" << code;
```
panda-sheep, line 115: A little suggestion,
Should I print the task name of the task into the log?

liuyu85cn: good catch, fixed

jievince (approved):
LGTM!

panda-sheep (approved):
Excellent!

PR #208, opened 2020-11-14
Encode string null in index as length in schema 

critical27:
Some bug fix and optimization of `IndexKeyUtils`:
1. null string will only encode as one byte, which make index scan not available (same as 1.0), for now we will encode as length of schema.
2. the length of `datetime` is wrong
3. we could support `lookup ... where tag.col == null` in storage
4. reduce one string copy in lookup

Besides, some code in `IndexKeyUtils` has been refactored, e.g. `collectIndexValues`, `vertexIndexKey`, `edgeIndexKey` and `encodeValues`. Previously we need an `colsType` to tell whether index has nullable column, which is unnecessary for now.

### DIFF
--- a/src/mock/MockData.cpp
+++ b/src/mock/MockData.cpp
@@ -799,12 +799,11 @@ std::vector<std::pair<PartitionID, std::string>> MockData::mockPlayerIndexKeys(b
             name = player.name_;
         }
         auto part = std::hash<std::string>()(name) % 6 + 1;
-        std::vector<Value> values;
-        values.emplace_back(encodeFixedStr(name, 20));
-        values.emplace_back(player.age_);
-        values.emplace_back(player.playing_);
-        std::vector<Value::Type> colsType;
-        auto key = IndexKeyUtils::vertexIndexKey(32, part, 1, name, values, colsType);
+        std::string values;
+        values.append(encodeFixedStr(name, 20));
+        values.append(IndexKeyUtils::encodeValue(player.age_));
+        values.append(IndexKeyUtils::encodeValue(player.playing_));
+        auto key = IndexKeyUtils::vertexIndexKey(32, part, 1, name, std::move(values));
         auto pair = std::make_pair(part, std::move(key));
         keys.emplace_back(std::move(pair));
     }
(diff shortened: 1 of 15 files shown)


src/storage/exec/IndexScanNode.h
```diff
@@ -100,56 +100,62 @@ class IndexScanNode : public RelNode<T> {
...
-                }
-                auto len = *iter->get_type().get_type_length();
-                prefix.append(IndexKeyUtils::encodeValue(col.get_begin_value(), len));
-            } else {
-                prefix.append(IndexKeyUtils::encodeValue(col.get_begin_value()));
+            auto type = IndexKeyUtils::toValueType(iter->type.type);
+            if (type == Value::Type::STRING && !iter->type.__isset.type_length) {
+                return Status::Error("String property index has not set prefix length.");
             }
+            prefix.append(encodeValue(col.begin_value, type, iter->type.get_type_length()));
```
panda-sheep, line 114: When col  is a not string type, ColumnTypeDef::type_length is not set. 
So the original code is reasonable.

src/storage/exec/IndexScanNode.h
```diff
@@ -100,56 +100,62 @@ class IndexScanNode : public RelNode<T> {
...
-                          ? *iter->get_type().get_type_length() : 0;
+            auto type = IndexKeyUtils::toValueType(iter->get_type().get_type());
+            if (type == Value::Type::STRING && !iter->get_type().__isset.type_length) {
+                return Status::Error("String property index has not set prefix length.");
+            }
             if (col.get_scan_type() == cpp2::ScanType::PREFIX) {
-                start.append(IndexKeyUtils::encodeValue(col.get_begin_value(), len));
-                end.append(IndexKeyUtils::encodeValue(col.get_begin_value(), len));
+                start.append(encodeValue(col.begin_value, type, iter->type.get_type_length()));
+                end.append(encodeValue(col.begin_value, type, iter->type.get_type_length()));
```
panda-sheep, line 138: ditto

src/utils/IndexKeyUtils.h
```diff
@@ -409,10 +408,7 @@ class IndexKeyUtils final {
             }
             offset += len;
         }
-        /*
-         * here need a string copy.
-         */
-        auto propVal = key.substr(offset, len);
+        auto propVal = key.subpiece(offset, len);
```
panda-sheep, line 411: line 397 need to handle

panda-sheep (commented):
Good job, excellent!

src/utils/IndexKeyUtils.h
critical27: Good catch

src/mock/MockData.cpp
```diff
@@ -903,17 +902,16 @@ std::vector<std::pair<PartitionID, std::string>> MockData::mockServeIndexKeys()
...
         auto part = std::hash<std::string>()(serve.playerName_) % 6 + 1;
-        std::vector<Value> values;
-        values.emplace_back(encodeFixedStr(serve.playerName_, 20));
-        values.emplace_back(encodeFixedStr(serve.teamName_, 20));
-        values.emplace_back(serve.startYear_);
+        std::string values;
+        values.append(encodeFixedStr(serve.playerName_, 20));
+        values.append(encodeFixedStr(serve.teamName_, 20));
+        values.append(IndexKeyUtils::encodeValue(serve.startYear_));
         std::vector<Value::Type> colsType;
```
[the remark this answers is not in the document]
critical27: 👌

src/storage/exec/IndexScanNode.h
panda-sheep: Got. In `IndexScanNode::encodeValue `， use it when the pointer `iter->type.get_type_length()` is valid

critical27: Yeah, it is a pointer.

panda-sheep (approved):
Well done！

bright-starry-sky (approved):
LGTM

PR #211, opened 2020-11-17
bugfix: FlushTask not return the correct error code type. 

liuyu85cn:
in FlushTask, if param spaceId is invalid. 
will return a kvstore::ResultCode

then use this(as an int, which is negative) to ctor a ErrorOr<cpp2::ErrorCode, std::vector<...>>

ResultCode will implicit as a really big size_t to ctor vector, 

which throw a bad_alloc exception. 

as the EitherOr class is noexcept, core.


liuyu85cn:
commit wrong

PR #212, opened 2020-11-17
fix FlushTask not return correct error code

liuyu85cn:
in FlushTask, if param spaceId is invalid. will return a kvstore::ResultCode
then use this(as an int, which is negative) to ctor a ErrorOr<cpp2::ErrorCode, std::vector<...>>
ResultCode will implicit as a really big size_t to ctor vector,
which throw a bad_alloc exception.
as the EitherOr class is noexcept, core.

also 

change type ResultCode from Enum to enum class, in case of any implicit error like this.

### DIFF
--- a/src/kvstore/Common.h
+++ b/src/kvstore/Common.h
@@ -16,7 +16,7 @@
 namespace nebula {
 namespace kvstore {
 
-enum ResultCode {
+enum class ResultCode {
     SUCCEEDED               = 0,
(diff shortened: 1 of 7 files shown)


dutor (approved):
Thanks !

panda-sheep (approved):
Excellent!

critical27 (approved):
Good job

PR #216, opened 2020-11-20
add getcolumn for context

nevermore3:
backgroud is https://www.yuque.com/nebulagraph/emchnb/kg2u5m
depend on vesoft-inc/nebula-common#316

### DIFF
--- a/src/utils/DefaultValueContext.h
+++ b/src/utils/DefaultValueContext.h
@@ -54,6 +54,11 @@ public:
         return Value::kEmpty;
     }
 
+    Value getColumn(int32_t) const override {
+        LOG(FATAL) << "Not allowed to call";
+        return Value::kEmpty;
+    }
+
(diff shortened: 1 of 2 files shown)


src/utils/DefaultValueContext.h
```diff
@@ -54,6 +54,11 @@ class DefaultValueContext final : public ExpressionContext {
         return Value::kEmpty;
     }
 
+    Value getColumn(int32_t) const override {
+        LOG(FATAL) << "Not allowed to call";
```
darionyaphet, line 58: `LOG(FATAL)` will take crash

nevermore3: emm. it's ok when vesoft-inc/nebula-common#316. merged

yixinglu (approved):

CPWstatic (approved):

yixinglu (approved):

PR #217, opened 2020-11-20
Support regular expression

jievince:
Depends on:
https://github.com/vesoft-inc/nebula-common/pull/287
https://github.com/vesoft-inc/nebula-graph/pull/414


dutor (commented):
Please add `TODO` here. Since this the regex match expression might be pushed down to the storage layer

jievince:
> Please add `TODO` here. Since this the regex match expression might be pushed down to the storage layer

Has added implementation  getRegex in StorageExpressionContext

dutor (approved):
LGTM.

PR #218, opened 2020-11-21
fix memory leak

critical27:
1. don't use `shared_from_this` in Listener, it seems has memory issue when `RaftPart` and `Host` has circular reference. I will modify one of them to be weak reference later if necessary.
2. remove useless code

### DIFF
--- a/src/kvstore/Listener.cpp
+++ b/src/kvstore/Listener.cpp
@@ -66,8 +66,7 @@ void Listener::start(std::vector<HostAddr>&& peers, bool) {
     // As for listener, we don't need Host actually. However, listener need to be aware of
     // membership change, it can be handled in preProcessLog.
     for (auto& addr : peers) {
-        auto hostPtr = std::make_shared<raftex::Host>(addr, shared_from_this());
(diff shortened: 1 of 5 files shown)


bright-starry-sky (approved):
Nice job.

critical27:
For example, this will report `Indirect memory leak`.

```
class Base;

class Member {
public:
    Member(std::shared_ptr<Base> ptr) : ptr_(ptr) {}
    std::shared_ptr<Base> ptr_;
};

class Base: public std::enable_shared_from_this<Base> {
public:
    virtual ~Base() {}

    virtual void init() {
        member_ = std::make_shared<Member>(shared_from_this());
    }

    std::shared_ptr<Member> member_;
};

class Derived: public Base {
public:
    virtual ~Derived() {}

    void init() override {
        member_ = std::make_shared<Member>(shared_from_this());
    }
};

TEST(test, test) {
    auto base = std::make_shared<Base>();
    base->init();

    auto derived = std::make_shared<Derived>();
    derived->init();
}
```

liuyu85cn (approved):
LGTM

PR #230, opened 2020-12-06
dedup result for index scan

bright-starry-sky:


Shylock-Hg:
Dedup in any case?

critical27:
Maybe we could add a dedup node, given some columns, dedup by these columns, and then return. Other method could use it as well.

bright-starry-sky:
> Maybe we could add a dedup node, given some columns, dedup by these columns, and then return. Other method could use it as well when more operator is pushed down.

good suggest. so we can push down yield clause . such as distinct.

Shylock-Hg:
> > Maybe we could add a dedup node, given some columns, dedup by these columns, and then return. Other method could use it as well when more operator is pushed down.
> 
> good suggest. so we can push down yield clause . such as distinct.

Yes, could add the `dedup` option in request like other interface like `getProps`.

critical27:
> > > Maybe we could add a dedup node, given some columns, dedup by these columns, and then return. Other method could use it as well when more operator is pushed down.
> > 
> > 
> > good suggest. so we can push down yield clause . such as distinct.
> 
> Yes, could add the `dedup` option in request like other interface like `getProps`.

The problem in `fetch` is that the first column is vid, the dedup result will be strange, that's why I insist remove the first column in previous PR. This won't effect `go`, because, go intends to dedup in the edge dataset.

Shylock-Hg:
> > > > Maybe we could add a dedup node, given some columns, dedup by these columns, and then return. Other method could use it as well when more operator is pushed down.
> > > 
> > > 
> > > good suggest. so we can push down yield clause . such as distinct.
> > 
> > 
> > Yes, could add the `dedup` option in request like other interface like `getProps`.
> 
> The problem in `fetch` is that the first column is vid, the dedup result will be strange, that's why I insist remove the first column in previous PR. This won't effect `go`, because, go intends to dedup in the edge dataset.

Yes, now IndexScan had no default return column. If we can provide the real vid in getProps it could be implicit too.

critical27:
> > > > > Maybe we could add a dedup node, given some columns, dedup by these columns, and then return. Other method could use it as well when more operator is pushed down.
> > > > 
> > > > 
> > > > good suggest. so we can push down yield clause . such as distinct.
> > > 
> > > 
> > > Yes, could add the `dedup` option in request like other interface like `getProps`.
> > 
> > 
> > The problem in `fetch` is that the first column is vid, the dedup result will be strange, that's why I insist remove the first column in previous PR. This won't effect `go`, because, go intends to dedup in the edge dataset.
> 
> Yes, now IndexScan had no default return column. If we can provide the real vid in getProps it could be implicit too.

Yeah, you guys should take whether we could remove the first column vid in `fetch` into consideration, IMO, it is worthy. When more operators are pushed to storage, the disadvantages will be more obvious.

bright-starry-sky:
Support DeDupNode.

src/storage/exec/IndexOutputNode.h
```diff
@@ -215,75 +198,152 @@ class IndexOutputNode final : public RelNode<T> {
...
         } else {
-            row.emplace_back(vId.subpiece(0, vId.find_first_of('\0')).toString());
+            row.emplace_back(std::move(v));
         }
+        return Status::OK();
     }
 
 private:
-    nebula::DataSet*                                  result_;
+    DeDupDataSet*                                     result_;
```
critical27, line 346: How about move the logic of dedup to `dedup node`, I think we can just remain the output, otherwise, other method which will try to use dedup need to process like what you did here.

src/storage/index/LookupBaseProcessor.inl
```diff
@@ -126,6 +129,10 @@ bool LookupBaseProcessor<REQ, RESP>::isOutsideIndex(Expression* filter,
  *              |  AggregateNode   |
  *              +--------+---------+
  *                       |
+ *              +--------+---------+
+ *              |    DeDupNode     |
```
Shylock-Hg, line 133: Need option in request to choose does dedup.

src/storage/exec/IndexOutputNode.h
bright-starry-sky: > How about move the logic of dedup to `dedup node`, I think we can just remain the output, otherwise, other method which will try to use dedup need to process like what you did here.

Emm... Actually, I've thought about it . this might OOM more likely. because if just keep the single output, we need to de-dup the data via logic` vector -> map -> vector` for all of result . 
WDYT? and any good idea?

src/storage/index/LookupBaseProcessor.inl
bright-starry-sky: yeah, I will add the distinct flag of thrift-api for lookup in the future.

src/storage/exec/IndexOutputNode.h
critical27: Make sense for dedup, for now, the logic is output `map<row, row>` and convert to `vector<row>` in dedup node. What I concern is that other operator will consume more memory than dedup, such as `order by` and `group by`, so maybe we can ignore OOM here. This could be controlled with max count for temp solution if necessary. And maybe we need to add something like `memtracker` to control memory in global.

BTW, if storage can't dedup and have a OOM, graph will OOM as well.

bright-starry-sky: > Make sense for dedup, for now, the logic is output `map<row, row>` and convert to `vector<row>` in dedup node. What I concern is that other operator will consume more memory than dedup, such as `order by` and `group by`, so maybe we can ignore OOM here. This could be controlled with max count for temp solution if necessary. And maybe we need to add something like `memtracker` to control memory in global.
> 
> BTW, if storage can't dedup and have a OOM, graph will OOM as well.

Make sense.

bright-starry-sky:
improved `DeDupNode`

critical27:
LGTM now, plz resolve the conflicts

bright-starry-sky:
resolved conflicts, and changed vid from string to int64_t.

critical27 (approved):
Good job

panda-sheep (approved):
Good job,
I have a doubt,
why not design it in the following way:
     *                    +--------+---------+
     *                    |  AggregateNode   |
     *                    +--------+---------+
     *                                   |
     *                    +--------+---------+
     *                    |    DeDupNode     |
     *                    +--------+---------+
     *                                   |
     *   +----------+-----------+     +----------+-----------+
     *   +   IndexOutputNode    +     +   IndexOutputNode    +
     *   +----------+-----------+     +----------+-----------+
     *                    |                                          | 
     *   +--------+-------------+     +--------+-------------+ 
     *   |       DeDupNode      |     |       DeDupNode      |
     *   +--------+-------------+     +--------+-------------+
     *                    |                                        |
     *   +----------+-----------+     +----------+-----------+
     *   +    IndexScanNode     +     +    IndexScanNode     +
     *   +----------+-----------+     +----------+-----------+

bright-starry-sky:
> Good job,
> I have a doubt,
> why not design it in the following way:
> 
> image

Aha，good question. since the value of each `outputNode` is already unique, we need to figure out the uniqueness of multiple `outputNodes`. So I think we don't need to deal with the uniqueness of each outputNode. WDYT?

PR #243, opened 2020-12-16
bug fix: shutdown storage server may core, if internal storage thread is not stopped. & some toss bug fix

liuyu85cn:
1. shutdown storage server may core, if internal storage thread is not stopped
2. toss bug fix:
    2.1 toss iter will not skip invalid resume lock
    2.2 toss iter won't wait all resume task finished
    2.3 revese key get wrong part in txn manager 

depends on https://github.com/vesoft-inc/nebula-common/pull/352


src/storage/transaction/TossEdgeIterator.h
```diff
@@ -167,24 +172,34 @@ class TossEdgeIterator : public SingleEdgeIterator {
...
+            LOG_IF(INFO, FLAGS_trace_toss) << "next(), waiting resume finished";
+            folly::collectAll(std::move(resumeTasks_)).wait();
             needWaitResumeTask_ = false;
             recoverEdgesIter_ = recoverEdges_.begin();
         } else {
             recoverEdgesIter_++;
         }
 
-        if (recoverEdgesIter_ != recoverEdges_.end()) {
+        while (recoverEdgesIter_ != recoverEdges_.end()) {
```
critical27, line 188: Good catch

src/storage/transaction/TransactionManager.cpp
```diff
@@ -185,14 +185,18 @@ folly::Future<cpp2::ErrorCode> TransactionManager::addSamePartEdges(
                     // steps 4 & 5: multi put local edges & multi remove persist locks
                     kvstore::BatchHolder bat;
                     for (auto& lock : lockData) {
-                        bat.remove(NebulaKeyUtils::toLockKey(lock.first));
+                        bat.remove(std::string(lock.first));
```
critical27, line 188: Maybe could move

src/storage/transaction/TransactionManager.cpp
```diff
@@ -391,7 +390,7 @@ folly::SemiFuture<kvstore::ResultCode> TransactionManager::commitEdgeOut(size_t
     if (!indexes.empty()) {
         std::vector<kvstore::KV> data;
         data.reserve(32);
-        data.emplace_back(std::move(key), std::move(props));
+        data.emplace_back(std::string(key), std::string(props));
```
critical27, line 393: Why we can't move here?

critical27 (commented):
Generally LGTM

src/storage/transaction/TransactionManager.cpp
liuyu85cn: fixed

src/storage/transaction/TransactionManager.cpp
liuyu85cn: fixed

critical27 (approved):

panda-sheep (approved):
excellent!

PR #245, opened 2020-12-17
fix fetch prop on tag not get the correct result

panda-sheep:
As title.
When there are multiple instances and multiple replicas, 
sometimes you get results from `fetch prop on tag`,  sometimes you don’t get results.
Single replica instance will not happen. 

chaos finds this problem，thanks chaos.


critical27 (approved):

liuyu85cn (approved):
LGTM

bright-starry-sky (approved):

PR #246, opened 2020-12-17
fix integer vId

critical27:
1. support int vId
2. fix storage can't be stopped
3. add uuid


src/storage/exec/IndexOutputNode.h
```diff
@@ -276,7 +276,7 @@ class IndexOutputNode final : public RelNode<T> {
 
     void addVidToRow(VertexIDSlice vId, Row& row) {
         if (planContext_->isIntId_) {
-            row.emplace_back(vId.toString());
+            row.emplace_back(*reinterpret_cast<const int64_t*>(vId.data()));
```
darionyaphet, line 279: So vid will be explained as a string previously?

darionyaphet (dismissed):

src/storage/exec/IndexOutputNode.h
critical27: Still need to copy, this is to prevent the id is more than expected 8 bytes.

panda-sheep (approved):
Good，LGTM

panda-sheep (approved):
LGTM

bright-starry-sky (approved):
Well done

nevermore3 (dismissed):

bright-starry-sky (approved):

panda-sheep (approved):

PR #248, opened 2020-12-18
Fixed index rebuilding task

dutor:
The index rebuilding task _keeps_ running and never finishes. The reasons are:
 * The rebuilding job is executed in a ThreadPoolExecutor.
 * The status is marked as RUNNING when a job is scheduled.
 * The job throws an exception during execution.
 * The ThreadPoolExecutor catches the exception and abort the job silently.
 * The status of the job has no chance to be updated.
 * Type of the exception is `std::length_error`, thrown by `std::string::apend(size_t, char)`.
 * The root cause is regarding a sized buffer as a null terminated string.


critical27 (approved):
Good catch

liuyu85cn (approved):

panda-sheep (approved):
👍

PR #252, opened 2020-12-22
bug fix: update may crash if use toss. 

liuyu85cn:
1. the key_ of update node will set in atomic op, so use a place holder key pass to transaction mgr
2. toss edge iterator have a stopSearching_ member variable which is dup with its parent class, change to another name.


src/storage/transaction/TransactionUtils.cpp
```diff
@@ -11,6 +11,10 @@
 namespace nebula {
 namespace storage {
 
+std::string TransactionUtils::dumpKey(const cpp2::EdgeKey& key) {
+    return folly::sformat("dumpKey(): src={}, dst={}", key.src.toString(), key.dst.toString());
```
darionyaphet, line 15: this is only used by test case?

critical27 (approved):
LGTM

src/storage/transaction/TossEdgeIterator.h
```diff
@@ -243,10 +248,10 @@ class TossEdgeIterator : public SingleEdgeIterator {
     /**
      * getNeighbors need to scan all edges, but update will stop at first edge,
      * use stopAtFirstEdge_ to let caller tell if this need to stop early
-     * use stopSearching_ to judge inside.
+     * use stopSearch_ to judge inside.
      */
     bool                                                 stopAtFirstEdge_{false};
-    bool                                                 stopSearching_{false};
+    bool                                                 stopSearch_{false};
```
critical27, line 254: Just delete this declare will work?

critical27 (approved):
LGTM

panda-sheep (approved):
LGTM

PR #260, opened 2020-12-23
Fix fetch problem

critical27:
Revert to previous behavior, this need to be fixed later because it involves one extra io, and it is weird.
For example, v1 don't have t1 and t2, but has t3. When fetch t1 and t2, it will be returned as one row as well.

Will change to #246 later.

https://github.com/vesoft-inc/nebula-graph/pull/507/checks?check_run_id=1600034306


nevermore3 (approved):

panda-sheep (approved):
Good!

PR #261, opened 2020-12-24
fix storage server stop, fix meta client related ut would hang

critical27:
1. Ut would hang if hostname does not parsed correctly.
2. Fix storage server stop reentrant
3. Delete the pragma ... see https://github.com/vesoft-inc/nebula-common/pull/365.


panda-sheep (approved):
LGTM

bright-starry-sky (approved):
Well done.

PR #266, opened 2020-12-28
Fix the integer vid type error.

Shylock-Hg:
The type error will lead the push-down expression computation error.


critical27 (approved):
Good catch

dutor (approved):

PR #269, opened 2020-12-29
fix admin client leader change and bug in transfer leader

critical27:
1. When leader change returned to AdminClient, the address is the data port, need to convert to admin port.
2. Introduced in #199, transfer leader should not return error in some case, actually the transfer succeed. This would incur balance leader/data failed
3. fix `show hosts meta` error which will return duplicate result.

### DIFF
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -417,6 +417,8 @@ void AdminClient::getResponse(std::vector<HostAddr> hosts,
                                         std::move(respGen));
                             return;
                         }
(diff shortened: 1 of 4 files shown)


bright-starry-sky (approved):
LGTM

panda-sheep (dismissed):
Good,  LGTM

critical27:
Hold on plz, there is another bug reported by thanos.

liuyu85cn (dismissed):
LGTM

liuyu85cn (approved):

bright-starry-sky (approved):
LGTM

PR #228, opened 2020-12-01
Support rebuild multi indexes

darionyaphet:

PR #275, opened 2020-12-30
Simplify rebuild index count, storage support rebuild multi index

critical27:
1. `onFlyingRequest_` is error prone, make it simple.
2. support rebuild multi index in storage. @darionyaphet, plz go on with #228


darionyaphet (approved):

src/storage/CommonUtils.h
```diff
@@ -66,6 +66,21 @@ class StorageEnv {
     }
 };
 
+class IndexCountWrapper {
```
darionyaphet, line 69: Nice

liuyu85cn (approved):

PR #232, opened 2020-12-08
Replace injector with gmock, support clean invalid balance plan

critical27:
1. Replace all injector in ut with gmock
2. Support clean invalid balance plan, rely on https://github.com/vesoft-inc/nebula-common/pull/342, https://github.com/vesoft-inc/nebula-graph/pull/463/files

PR #282, opened 2020-12-31
Fix some issue in balance data

critical27:
Some fix during balance test:
1. Balance plan is stored out of order, which may cause recover and reset an outdate plan. For now it will stored in timestamp descend order.
2. All steps in a Balance Task could be executed more than once, in case we recover a failed plan.
3. Remove some redundant code of balance utils introduced in #199 
4. Revert the logic of INVALID task to previous behavior introduced in #232
    * When a plan is stopped or dst is not alive any more, a task will be marked as INVALID, the task will not be executed again. Balancer will start a new plan instead.
    * Only recover failed task, skip the in-progress and invalid tasks


src/meta/processors/admin/BalancePlan.cpp
```diff
@@ -184,8 +184,10 @@ cpp2::ErrorCode BalancePlan::recovery(bool resume) {
             task.startTimeMs_ = std::get<2>(tup);
             task.endTimeMs_ = std::get<3>(tup);
             if (resume && task.ret_ != BalanceTaskResult::SUCCEEDED) {
-                // Resume any task not finished
-                task.ret_ = BalanceTaskResult::IN_PROGRESS;
+                // Resume the failed task, skip the in-progress and invalid tasks
+                if (task.ret_ == BalanceTaskResult::FAILED) {
```
darionyaphet, line 188: use to avoid cover invalid task ？

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -61,10 +64,12 @@ void BalanceTask::invoke() {
             if (srcLived) {
                 client_->transLeader(spaceId_, partId_, src_).thenValue([this](auto&& resp) {
                     if (!resp.ok()) {
-                        LOG(ERROR) << taskIdStr_ << "Transfer leader failed, status " << resp;
                         if (resp == nebula::Status::PartNotFound()) {
-                            ret_ = BalanceTaskResult::INVALID;
+                            // if the partition has been removed before, regard as succeeded
+                            LOG(WARNING) << "Can't find part " << partId_ << " on " << src_;
```
darionyaphet, line 69: ERROR?

src/meta/processors/admin/BalancePlan.cpp
critical27: Yeah, this is the behavior before.

src/meta/processors/admin/BalanceTask.cpp
critical27: It is not an error actually, it could happen when one part has been moved brfore.

src/meta/test/BalancerTest.cpp
```diff
@@ -1174,16 +1145,9 @@ TEST(BalanceTest, CleanLastInvalidBalancePlanTest) {
     CHECK(ok(ret));
     auto balanceId = value(ret);
 
-    sleep(1);
-    // stop the running plan
-    TestUtils::registerHB(kv.get(), {{"0", 0}, {"1", 1}, {"2", 2}});
-    auto stopRet = balancer.stop();
```
darionyaphet, line 1180: why don't check the stop result ?

darionyaphet (approved):

src/meta/test/BalancerTest.cpp
critical27: Previously the task is delayed by 3 seconds, so need to stop the others task. This ut has been changed, so no need to stop anymore.

panda-sheep (approved):
LGTM

Issue #270, opened 2020-12-29
balance leader core


critical27:
![image](https://user-images.githubusercontent.com/13706157/103264526-ebf56b00-49e5-11eb-95be-bf8ed1e572f2.png)

critical27:
Found, will fix later

panda-sheep:
fixed in pr 290

PR #290, opened 2021-01-08
fix map erase

panda-sheep:
As title.
Thanks chaos for checking out this problem.

### DIFF
--- a/src/meta/processors/admin/Balancer.cpp
+++ b/src/meta/processors/admin/Balancer.cpp
@@ -738,11 +738,13 @@ int32_t Balancer::giveupLeaders(HostParts& leaderParts,
     int32_t taskCount = 0;
     auto& sourceLeaders = leaderParts[source];
     size_t maxLoad = hostBounds_[source].second;
+
     // host will try to transfer the extra leaders to other peers
-    for (auto it = sourceLeaders.begin(); it != sourceLeaders.end(); it++) {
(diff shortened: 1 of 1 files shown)


critical27 (commented):
image

Good job, I found this as well, we don't need `isErase` actually. Close #270

panda-sheep:
> image
> 
> Good job, I found this as well, we don't need `isErase` actually. Close #270

Yes, this kind of reform is also ok，  see https://github.com/vesoft-inc/nebula-cpp/pull/8

critical27 (approved):

src/meta/processors/admin/Balancer.cpp
```diff
@@ -752,25 +754,33 @@ int32_t Balancer::giveupLeaders(HostParts& leaderParts,
...
-            it = sourceLeaders.erase(it);
-            targetLeaders.emplace_back(partId);
-            plan.emplace_back(spaceId, partId, source, *target);
-            LOG(INFO) << "giveup plan trans leader space: " << spaceId
-                      << " part: " << partId
-                      << " from " << source.host << ":" << source.port
-                      << " to " << target->host << ":" << target->port;
-            ++taskCount;
+        if (target != targets.end()) {
+            auto& targetLeaders = leaderParts[*target];
```
liuyu85cn, line 760: I'm not sure if leaderParts[*target] may trigger re-hash, and invalid everything

liuyu85cn (approved):
good job.

PR #291, opened 2021-01-08
fixed the bug for restore meta.

monadbobo:
1. Fix the bug of incorrectly calling storage's interface when restore meta (no need to copy to follower node).

2. add the interface to write storage directly to Nebulastore.

### DIFF
--- a/src/kvstore/KVStore.h
+++ b/src/kvstore/KVStore.h
@@ -208,6 +208,9 @@ public:
     virtual ResultCode restoreFromFiles(GraphSpaceID spaceId,
                                         const std::vector<std::string>& files) = 0;
 
+    virtual ResultCode multiPutWitoutReplicator(GraphSpaceID spaceId,
+                                                std::vector<KV> keyValues) = 0;
+
 protected:
     KVStore() = default;
 };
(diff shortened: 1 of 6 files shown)


src/kvstore/KVStore.h
```diff
@@ -208,6 +208,9 @@ class KVStore {
     virtual ResultCode restoreFromFiles(GraphSpaceID spaceId,
                                         const std::vector<std::string>& files) = 0;
 
+    virtual ResultCode multiPutWitoutReplicator(GraphSpaceID spaceId,
```
bright-starry-sky, line 211: `multiPutWithoutReplicator`

critical27:
Em, I am not pretty sure if this PR will effect other function. As far as I know, the http replace interface do need to write to all meta.

monadbobo:
> Em, I am not pretty sure if this PR will effect other function. As far as I know, the http replace interface do need to write to all meta.

The current http interface does not use replaceHostInPartition, so perhaps we can add a parameter to replaceHostInPartition to control whether it needs to take the draft protocol.

critical27 (approved):

panda-sheep (approved):
Good，Excellent！

PR #292, opened 2021-01-11
Task report finish to job manager

liuyu85cn:
What problem this pr want to address: 

when JobManager(meta) dispatch tasks to storage, 
it will call "collectAll" to wait all task finished.
if meta leader change during this collectAll, 
meta will fail to write job status to kvstore, forever.

So, don't let meta wait for those tasks. Instead of when a single task finished, 
task manager(storage) will call reportTaskFinish to the leader of meta.

### DIFF
--- a/src/meta/processors/jobMan/JobManager.cpp
+++ b/src/meta/processors/jobMan/JobManager.cpp
@@ -255,7 +255,6 @@ cpp2::ErrorCode JobManager::jobFinished(JobID jobId, bool suc) {
     cleanJob(jobId);
 }
 
-// cpp2::ErrorCode JobManager::reportTaskFinish(JobID jobId, int32_t taskId, cpp2::ErrorCode code) {
 cpp2::ErrorCode JobManager::reportTaskFinish(const cpp2::ReportTaskReq& req) {
     auto jobId = req.get_job_id();
     auto taskId = req.get_task_id();
@@ -284,6 +283,7 @@ cpp2::ErrorCode JobManager::reportTaskFinish(const cpp2::ReportTaskReq& req) {
     auto taskStatus =
         code == cpp2::ErrorCode::SUCCEEDED ? cpp2::JobStatus::FINISHED : cpp2::JobStatus::FAILED;
     task->setStatus(taskStatus);
+    save(task->taskKey(), task->taskVal());
 
     auto allTaskFinished = std::none_of(tasks.begin(), tasks.end(), [](auto& task){
         return task.status_ == cpp2::JobStatus::RUNNING;
--- a/src/meta/processors/jobMan/JobManager.cpp
+++ b/src/meta/processors/jobMan/JobManager.cpp
@@ -198,11 +198,11 @@ bool JobManager::runJobInternalOld(const JobDescription& jobDesc) {
 // @return: true if all task dispatched, else false
 bool JobManager::runJobInternal(const JobDescription& jobDesc) {
     currJob_ = MetaJobExecutorFactory::createMetaJobExecutor(jobDesc, kvStore_, adminClient_);
-
     if (currJob_ == nullptr) {
         LOG(ERROR) << "unreconized job cmd " << static_cast<int>(jobDesc.getCmd());
         return false;
     }
+
     if (jobDesc.getStatus() == cpp2::JobStatus::STOPPED) {
         currJob_->stop();
         return true;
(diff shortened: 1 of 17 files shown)


src/meta/processors/jobMan/JobManager.cpp
```diff
@@ -79,10 +79,10 @@ void JobManager::shutDown() {
 
 void JobManager::scheduleThread() {
     LOG(INFO) << "JobManager::runJobBackground() enter";
-    while (status_ == Status::RUNNING) {
+    while (status_ != JbmgrStatus::STOPPED) {
```
panda-sheep, line 82: why not while (status_ == JbmgrStatus::IDLE || status_ == JbmgrStatus::BUSY) ？

src/meta/processors/jobMan/JobManager.cpp
```diff
@@ -79,10 +79,10 @@ void JobManager::shutDown() {
...
         int32_t iJob = 0;
-        while (!try_dequeue(iJob)) {
-            if (status_ == Status::STOPPED) {
+        while (status_ == JbmgrStatus::BUSY || !try_dequeue(iJob)) {
```
panda-sheep, line 84: the status_ here can be IDLE?

panda-sheep (commented):
Good job!

src/meta/processors/jobMan/JobManager.cpp
panda-sheep: my fault you are right

src/meta/processors/jobMan/StatisJobExecutor.cpp
```diff
@@ -57,6 +57,56 @@ StatisJobExecutor::executeInternal(HostAddr&& address, std::vector<PartitionID>&
                                  std::move(parts), concurrency_, &(statisItem_[address]));
 }
 
+void showStatisItem(const cpp2::StatisItem& item, const std::string& msg) {
```
panda-sheep, line 60: What is this interface for?

src/meta/processors/jobMan/StatisJobExecutor.cpp
```diff
@@ -57,6 +57,56 @@ StatisJobExecutor::executeInternal(HostAddr&& address, std::vector<PartitionID>&
...
+ */
+cpp2::ErrorCode StatisJobExecutor::saveSpecialTaskStatus(const cpp2::ReportTaskReq& req) {
+    if (!req.__isset.statis) {
+        return cpp2::ErrorCode::SUCCEEDED;
+    }
+    cpp2::StatisItem statisItem;
+    auto statisKey = MetaServiceUtils::statisKey(space_);
+    std::string val;
+    auto ret = kvstore_->get(kDefaultSpaceId, kDefaultPartId, statisKey, &val);
+    if (ret == kvstore::ResultCode::SUCCEEDED) {
```
panda-sheep, line 101: There maybe be a problem here:
If the last statis job and the current statis job are in the same space,
do you add the results of the last time?

src/meta/processors/jobMan/StatisJobExecutor.cpp
liuyu85cn: for debug ... and I think it maybe used one day. If you don't like it, I'm ok to remove those.

src/meta/processors/jobMan/StatisJobExecutor.cpp
liuyu85cn: discussed offline, The previous status value will be init at prepare state.

src/meta/processors/jobMan/JobManager.cpp
```diff
@@ -91,78 +85,210 @@ void JobManager::scheduleThread() {
...
+    auto jobExec =
+        MetaJobExecutorFactory::createMetaJobExecutor(*optJobDesc, kvStore_, adminClient_);
+
+    if (jobExec) {
+        if (!optJobDesc->getParas().empty()) {
+            auto spaceName = optJobDesc->getParas().front();
+            auto spaceId = getSpaceId(spaceName);
+            LOG(INFO) << folly::sformat("spaceName={}, spaceId={}", spaceName, spaceId);
+            if (spaceId != -1) {
+                jobExec->setSpaceId(spaceId);
```
critical27, line 174: Maybe we could add a spaceId in thrift `AdminJobReq`, then the paras could just save their own parameters later. And job description could directly use it when job/task added.

src/meta/processors/jobMan/StatisJobExecutor.cpp
```diff
@@ -57,51 +64,83 @@ StatisJobExecutor::executeInternal(HostAddr&& address, std::vector<PartitionID>&
...
+
+    lhs.space_vertices += rhs.space_vertices;
+    lhs.space_edges += rhs.space_edges;
+    // ignore part_corelativity, they shoule be same
+}
+
+/**
+ * @brief caller will guarantee there won't be any conflict read / write.
+ */
+cpp2::ErrorCode StatisJobExecutor::saveSpecialTaskStatus(const cpp2::ReportTaskReq& req) {
```
critical27, line 100: Do we need to add lock for accumulate result?

src/meta/processors/jobMan/JobManager.h
```diff
@@ -113,13 +134,13 @@ class JobManager : public nebula::cpp::NonCopyable, public nebula::cpp::NonMovab
     // The job in running or queue
     folly::ConcurrentHashMap<JobID, JobDescription>    inFlightJobs_;
 
-    std::unique_ptr<thread::GenericWorker>             bgThread_;
+    std::thread                                        bgThread_;
```
critical27, line 137: What is the issue of `GenericWorker`?

src/meta/processors/jobMan/StatisJobExecutor.cpp
liuyu85cn: no, like I said in comments.

critical27: 👌

src/meta/processors/jobMan/JobManager.h
liuyu85cn: got some weird dead lock. And I believe I just need a simple thread to run "scheduleThread".

critical27: 👌

src/meta/processors/jobMan/JobManager.cpp
liuyu85cn: ok, and I think in another pr is better, in which I can change all the existing job

src/meta/processors/jobMan/JobManager.cpp
liuyu85cn: because it's long ...

critical27 (approved):
LGTM

bright-starry-sky (approved):
LGTM

PR #295, opened 2021-01-11
Support multi rebuild index

darionyaphet:


src/storage/admin/RebuildIndexTask.cpp
```diff
@@ -18,16 +18,23 @@ RebuildIndexTask::genSubTasks() {
     space_ = ctx_.parameters_.space_id;
     auto parts = ctx_.parameters_.parts;
 
-    auto itemsRet = getIndexes(space_);
-    if (!itemsRet.ok()) {
-        LOG(ERROR) << "Indexes Not Found";
+    if (ctx_.parameters_.task_specfic_paras.empty()) {
```
critical27, line 21: How about build all index when empty?

laura-ding: good idea

darionyaphet: ok

src/storage/admin/RebuildIndexTask.cpp
```diff
@@ -18,15 +18,32 @@ RebuildIndexTask::genSubTasks() {
...
+                return cpp2::ErrorCode::E_INDEX_NOT_FOUND;
+            }
+            items.emplace_back(indexRet.value());
+        }
     }
 
-    auto items = itemsRet.value();
     if (items.empty()) {
-        LOG(INFO) << "Index is empty";
+        LOG(ERROR) << "Index is empty";
```
bright-starry-sky, line 46: LOG(WARNING) ?

bright-starry-sky (approved):
LGTM

src/storage/admin/RebuildIndexTask.cpp
darionyaphet: the index id not found seems Error maybe better

liuyu85cn (approved):

PR #296, opened 2021-01-12
Fix raft bugs

critical27:
1. cherry pick https://github.com/vesoft-inc/nebula/pull/2419
2. Follower reset the timeout timer again in case wal and commit takes longer time than expected

Other changes need to be verified carefully, will open another PR to improve leader stability.


bright-starry-sky (approved):
Well done

darionyaphet (approved):

PR #298, opened 2021-01-14
Modify to support to rebuild multi indexes

laura-ding:


src/storage/test/RebuildIndexTest.cpp
```diff
@@ -75,7 +75,8 @@ TEST_F(RebuildIndexTest, RebuildTagIndexCheckALLData) {
     parameter.set_space_id(1);
     std::vector<PartitionID> parts = {1, 2, 3, 4, 5, 6};
     parameter.set_parts(parts);
-    parameter.task_specfic_paras.emplace_back("4,5");
+    parameter.task_specfic_paras.emplace_back("4");
```
darionyaphet, line 78: IF we will add more argument about rebuild how to determine the index id bounds ？

laura-ding: Can you tell me what parameters will be supported?  I think every parameter in the vector means the same meaning. so the user  use it is easy. and it no need to tell user the firstly parameter mean what, the second mean what.

critical27:
Better to wait #292 merge at first.

src/meta/processors/jobMan/ListEdgeIndexStatusProcessor.cpp
```diff
@@ -52,7 +52,13 @@ void ListEdgeIndexStatusProcessor::process(const cpp2::ListIndexStatusReq& req)
...
     std::unordered_map<std::string, cpp2::JobStatus> tmp;
     for (auto &jobDesc : jobs) {
-        tmp.emplace(jobDesc.get_paras()[0], jobDesc.get_status());
+        auto paras = jobDesc.get_paras();
+        if (paras.size() == 1) {
+            tmp.emplace(paras[0] + "_all_edge_indexes", jobDesc.get_status());
+            continue;
+        }
+        paras.pop_back();
+        tmp.emplace(folly::join(",", paras), jobDesc.get_status());
```
bright-starry-sky, line 61: In graph layer，I saw that null argument is allowed, meaning that all indexes. How to unify with the storage layer?
https://github.com/vesoft-inc/nebula-graph/pull/600，

laura-ding: The same goes for Storage.You can see the RebuildJobExecutor.cpp

bright-starry-sky: > The same goes for Storage.You can see the RebuildJobExecutor.cpp

Got, We do that in code `RebuildIndexTask.cpp`. nice job.

bright-starry-sky (approved):
LGTM

src/storage/test/RebuildIndexTest.cpp
darionyaphet: There are hidden dangers here

darionyaphet (approved):

PR #310, opened 2021-01-19
fix read readerV1 data from cache

critical27:
When we reset reader as RowReaderV1, it will trim the header, and we need to insert raw data into vertex cache instead of the trimmed one.

https://jira.nebula-graph.io/projects/NG/issues/NG-395


bright-starry-sky (approved):
LGTM

panda-sheep (approved):
Good job！

PR #318, opened 2021-01-21
Using executor pool for all scan interfaces

bright-starry-sky:


src/storage/index/LookupProcessor.cpp
```diff
@@ -10,6 +10,15 @@ namespace nebula {
 namespace storage {
 
 void LookupProcessor::process(const cpp2::LookupIndexRequest& req) {
+    if (executor_ != nullptr) {
+        executor_->add([req, this] () {
```
critical27, line 14: Maybe we could move the req instead of copy it?

bright-starry-sky: > Maybe we could move the req instead of copy it?

Good point. changed to `by-reference capture` for lambda capture. But I'm not sure if I'll be released by `GraphStorageServiceHandler` . WDYT?

critical27: I think capture by reference is ok.
image

critical27 (approved):

panda-sheep (approved):
well done！

src/storage/index/LookupProcessor.cpp
bright-starry-sky: Got, Thanks.

PR #319, opened 2021-01-21
performance optimize, if disable mvcc, may ignore some storage iterator check.

liuyu85cn:
as title

toss iterator is much complex.. not finished yet. 

for toss, looks like we can't get rid of lastRank_ and lastRank_.


critical27 (approved):
LGTM

panda-sheep (approved):
well done

PR #320, opened 2021-01-21
fix multi meta upgrade

critical27:
When multiple meta upgrade:
* leader will upgrade all data from v1 to v2
* follower just wait from leader

https://jira.nebula-graph.io/browse/NG-401

### DIFF
--- a/src/meta/MetaVersionMan.cpp
+++ b/src/meta/MetaVersionMan.cpp
@@ -15,46 +15,29 @@ DEFINE_bool(print_info, false, "enable to print the rewrite data");
 
 namespace nebula {
 namespace meta {
-static const int32_t metaVersion = 2;
+
 static const std::string kMetaVersionKey = "__meta_version__";          // NOLINT
+
 // static
-int32_t MetaVersionMan::getMetaVersionFromKV(kvstore::KVStore* kv) {
+MetaVersion MetaVersionMan::getMetaVersionFromKV(kvstore::KVStore* kv) {
     CHECK_NOTNULL(kv);
     std::string value;
     auto code = kv->get(kDefaultSpaceId, kDefaultPartId, kMetaVersionKey, &value, true);
-    if (code == kvstore::ResultCode::ERR_KEY_NOT_FOUND) {
-        LOG(INFO) << "There is no meta version key existed in kvstore!";
-        if (isV1(kv)) {
-            return 1;
-        } else {
-            auto ret = setMetaVersionToKV(kv);
-            if (!ret) {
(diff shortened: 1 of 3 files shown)


src/meta/MetaVersionMan.h
```diff
@@ -14,21 +14,28 @@
 namespace nebula {
 namespace meta {
 
+enum class MetaVersion {
+    UNKNOWN = 0,
```
laura-ding, line 18: why modify to enum, it's hard to expand in the future.

src/meta/MetaVersionMan.cpp
```diff
@@ -330,8 +313,13 @@ Status MetaVersionMan::doUpgrade(kvstore::KVStore* kv) {
             }
         }
     }
-    return Status::OK();
+    if (!setMetaVersionToKV(kv)) {
```
laura-ding, line 316: Not only set the version when upgrade. when the meta has not set the version also need to set the version

panda-sheep (approved):
Well done！

src/meta/MetaVersionMan.cpp
laura-ding: the 2.0 data couldn't call doUpgrade, the doUpgrade directly to decode by v1

critical27: OK, I will move it to daemon as before.

src/meta/MetaVersionMan.cpp
```diff
@@ -15,66 +15,49 @@ DEFINE_bool(print_info, false, "enable to print the rewrite data");
...
+        return MetaVersion::UNKNOWN;
     }
     if (iter->valid()) {
         auto v1KeySize = hostPrefix.size() + sizeof(int64_t);
-        if (iter->key().size() != v1KeySize) {
-            return false;
+        if (iter->key().size() == v1KeySize) {
+            return MetaVersion::V1;
+        } else {
+            return MetaVersion::V2;
```
bright-starry-sky, line 47: A little confusion here, is better as below ?
`return (iter->key().size() == v1KeySize) ? MetaVersion::V1 : MetaVersion::UNKNOWN;`

bright-starry-sky (dismissed):
Well done. LGTM.

src/meta/MetaVersionMan.cpp
critical27: Make sense.

bright-starry-sky (approved):

panda-sheep (approved):
well done！

PR #325, opened 2021-01-22
fixed error of  captures copy

bright-starry-sky:
Please hold this PR. a complete test is required.

### DIFF
--- a/src/storage/index/LookupProcessor.cpp
+++ b/src/storage/index/LookupProcessor.cpp
@@ -11,7 +11,7 @@ namespace storage {
 
 void LookupProcessor::process(const cpp2::LookupIndexRequest& req) {
     if (executor_ != nullptr) {
-        executor_->add([req, this] () {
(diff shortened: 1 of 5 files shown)


src/storage/index/LookupProcessor.cpp
```diff
@@ -11,7 +11,7 @@ namespace storage {
 
 void LookupProcessor::process(const cpp2::LookupIndexRequest& req) {
     if (executor_ != nullptr) {
-        executor_->add([req, this] () {
+        executor_->add([this, req = std::move(req)] () mutable {
```
Shylock-Hg, line 14: Move the constant variable is useless.

bright-starry-sky: > Move the constant variable is useless.

image
I believe that in all things there is a purpose, not ready to review. hold this PR.

bright-starry-sky:
This change will be discarded because passed a reference will cause `ThriftCancelError`

PR #328, opened 2021-01-25
Upgrade the meta index info.

Shylock-Hg:
NG-417

### DIFF
--- a/src/meta/upgradeData/MetaDataUpgrade.cpp
+++ b/src/meta/upgradeData/MetaDataUpgrade.cpp
@@ -109,7 +109,7 @@ Status MetaDataUpgrade::rewriteIndexes(const folly::StringPiece &key,
         schemaId.set_edge_type(oldItem.get_schema_id().getType());
     }
     newItem.set_schema_id(schemaId);
-    NG_LOG_AND_RETURN_IF_ERROR(convertToNewColumns(oldItem.fields, newItem.fields));
+    NG_LOG_AND_RETURN_IF_ERROR(convertToNewIndexColumns(oldItem.fields, newItem.fields));
     NG_LOG_AND_RETURN_IF_ERROR(put(key, MetaServiceUtils::indexVal(newItem)));
     return Status::OK();
 }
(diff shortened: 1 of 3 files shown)


Shylock-Hg:
Now the index meta info will upgrade to 2.0, and the new index data will write to separate place from 1.0 index data by rebuild.

bright-starry-sky:
> Now the index meta info will upgrade to 2.0, and the new index data will write to separate place from 1.0 index data by rebuild.
I found an additional problem, do we need to fix it in this PR?
for method `MetaDataUpgrade::rewriteIndexes`, if the field type is `STRING`, we need convert it from `STRING` to `FIXED_STRING(256)`.

Shylock-Hg:
> > Now the index meta info will upgrade to 2.0, and the new index data will write to separate place from 1.0 index data by rebuild.
> > I found an additional problem, do we need to fix it in this PR?
> > for method `MetaDataUpgrade::rewriteIndexes`, if the field type is `STRING`, we need convert it from `STRING` to `FIXED_STRING(256)`.

Which subversion of 1.x limit string index to 256 byte length?

Shylock-Hg:
> > Now the index meta info will upgrade to 2.0, and the new index data will write to separate place from 1.0 index data by rebuild.
> > I found an additional problem, do we need to fix it in this PR?
> > for method `MetaDataUpgrade::rewriteIndexes`, if the field type is `STRING`, we need convert it from `STRING` to `FIXED_STRING(256)`.

Fixed.

src/meta/upgradeData/MetaDataUpgrade.cpp
```diff
@@ -251,6 +251,29 @@ Status MetaDataUpgrade::convertToNewColumns(const std::vector<oldmeta::cpp2::Col
...
+            type.set_type(cpp2::PropertyType::FIXED_STRING);
+            type.set_type_length(256);
+            columnDef.set_type(std::move(type));
+        } else {
+            columnDef.type.set_type(static_cast<cpp2::PropertyType>(colDef.get_type().get_type()));
+        }
+        DCHECK(!colDef.__isset.default_value);
+        if (FLAGS_null_type) {
+            columnDef.set_nullable(true);
+        }
```
bright-starry-sky, line 271: Do we really need this nullable for ColumnDef? How is this parameter set for tag|edge schema ?

Shylock-Hg: I don't sure that is the index require the null flag so I don't modify it. If it's useless I could remove it.

bright-starry-sky: > I don't sure that is the index require the null flag so I don't modify it. If it's useless I could remove it.

got, please wait a moment. let me think about it.

bright-starry-sky: > > I don't sure that is the index require the null flag so I don't modify it. If it's useless I could remove it.
> 
> got, please wait a moment. let me think about it.

I think, because v1 doesn't support nullable parameter, when  the meta upgraded from v1 to v2,  we don't need to set nullable param to true(default value `false` is enough), so I think the ColumnDef of IndexItem also don't need set it. WDYT?

Shylock-Hg: In fact, 1.0 has no null but 2.0 has it.So Laura add FLAGS_null_type to determine true or false for the null option in 2.0 by user. And for index I reserve this logical if the index require the null option or I could remove this logical.

critical27: IMO, both add FLAGS_null_type in tag/edge and index schema, or neither of them add the flag.

critical27 (approved):
LGTM

src/meta/upgradeData/MetaDataUpgrade.cpp
bright-starry-sky: > In fact, 1.0 has no null but 2.0 has it.So Laura add FLAGS_null_type to determine true or false for the null option in 2.0 by user. And for index I reserve this logical if the index require the null option or I could remove this logical if index process don't use it.

Got, I've learned. keep it.

Shylock-Hg: Yes as @critical27 said, but if the null is useless in index process we could don't set it.

bright-starry-sky (approved):

PR #330, opened 2021-01-25
upgrade prepare check for index

bright-starry-sky:


panda-sheep (approved):
Excellent！

bright-starry-sky:
This time is too long, resulting in a mismatch with the current code, which will be solved through a new PR if necessary

PR #331, opened 2021-01-25
Improve delete edges, remove extra prefix in fetch

critical27:
1. Remove extra io in fetch. For example, v1 don't have t1 and t2, but has t3. When fetch t1 and t2, it will be returned as one row as well. 

**This will break graph ci.**, please fix it. @yixinglu @CPWstatic @nevermore3 @Shylock-Hg
  
  This is the previous ci log https://github.com/vesoft-inc/nebula-graph/pull/507/checks?check_run_id=1600034306.

`go ... yield edge.xxx, $$.xxx` will reproduce this problem as well, because no more `$$.xxx` returned. (previously will return an empty value if vertex has another tag).

2. Improve delete edge

**Don't merge plz, wait graph ready.**

### DIFF
--- a/src/storage/exec/GetPropNode.h
+++ b/src/storage/exec/GetPropNode.h
@@ -28,23 +28,16 @@ public:
         , vertexCache_(vertexCache) {}
 
     kvstore::ResultCode execute(PartitionID partId, const VertexID& vId) override {
-        // todo(doodle): revert to previous behavior, here is an extra io, will be deleted asap
-        // check is vertex exists
-        std::unique_ptr<kvstore::KVIterator> iter;
-        auto prefix = NebulaKeyUtils::vertexPrefix(planContext_->vIdLen_, partId, vId);
-        auto ret = planContext_->env_->kvstore_->prefix(planContext_->spaceId_,
-                                                        partId, prefix, &iter);
+        auto ret = RelNode::execute(partId, vId);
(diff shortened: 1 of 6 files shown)


src/storage/exec/GetPropNode.h
```diff
@@ -28,23 +28,16 @@ class GetTagPropNode : public QueryNode<VertexID> {
         , vertexCache_(vertexCache) {}
 
     kvstore::ResultCode execute(PartitionID partId, const VertexID& vId) override {
-        // todo(doodle): revert to previous behavior, here is an extra io, will be deleted asap
-        // check is vertex exists
-        std::unique_ptr<kvstore::KVIterator> iter;
-        auto prefix = NebulaKeyUtils::vertexPrefix(planContext_->vIdLen_, partId, vId);
```
Shylock-Hg, line 34: I think you could put them to another PR. And according to current behavior in graph we need to determine does the vertex exists, so after this PR we need to seek all tag in current space for this @vesoft-inc/graph-reviewers

critical27: If graph need to determine does the vertex exists, could call fetch *, as for those tags don't care, pass kTag. 

As the example said, v1 don't have t1 and t2, now we fetch t1 and t2. Current logic you can't explain why storage return different result.

nevermore3 (approved):

CPWstatic (approved):

PR #332, opened 2021-01-25
const VertexID&

critical27:
Modify all `VertexID` to `const VertexID&` in utils since we treat vid as string, avoid extra string copy.


bright-starry-sky (approved):
LGTM

panda-sheep (approved):
Good.

PR #342, opened 2021-02-01
Refactored the stats

sherman-the-tank:
1) Remove the dependency on stats::Stats
2) Now each individual processor can have `specific counters


critical27 (commented):
Generally LGTM, wait https://github.com/vesoft-inc/nebula-common/pull/392

src/storage/index/LookupProcessor.h
```diff
@@ -13,15 +13,17 @@
 namespace nebula {
 namespace storage {
 
+extern ProcessorCounters kLookupCounters;
```
critical27, line 16: Why not define all counters in `GraphStorageServiceHandler`?

sherman-the-tank: Each individual processor could have it own counters. So it makes more sense to let the processors own the counters

panda-sheep (approved):
Good job, LGTM

critical27 (approved):
Well done

PR #344, opened 2021-02-03
Use absolute path for nebula binary and config file in docker image

yixinglu:
as title.


jievince (approved):

darionyaphet:
If I specify that the installation directory will be affected?

bright-starry-sky (approved):
LGTM

PR #345, opened 2021-02-03
Fix snapshot promise

critical27:
1. Fix snapshot promise would be set more than once after vertex/edge separate.
2. Fix missing counters in `src/storage/GraphStorageServiceHandler.cpp`

### DIFF
--- a/src/storage/GraphStorageServiceHandler.cpp
+++ b/src/storage/GraphStorageServiceHandler.cpp
@@ -89,14 +89,14 @@ GraphStorageServiceHandler::future_updateVertex(const cpp2::UpdateVertexRequest&
 // Edge section
 folly::Future<cpp2::ExecResponse>
 GraphStorageServiceHandler::future_addEdges(const cpp2::AddEdgesRequest& req) {
-    auto* processor = AddEdgesProcessor::instance(env_);
+    auto* processor = AddEdgesProcessor::instance(env_, &kAddEdgesCounters);
     RETURN_FUTURE(processor);
 }
 
 
 folly::Future<cpp2::ExecResponse>
 GraphStorageServiceHandler::future_deleteEdges(const cpp2::DeleteEdgesRequest& req) {
-    auto* processor = DeleteEdgesProcessor::instance(env_);
+    auto* processor = DeleteEdgesProcessor::instance(env_, &kDelEdgesCounters);
     RETURN_FUTURE(processor);
 }
 
@@ -164,7 +164,7 @@ GraphStorageServiceHandler::future_getUUID(const cpp2::GetUUIDReq&) {
 
 folly::Future<cpp2::ExecResponse>
 GraphStorageServiceHandler::future_addEdgesAtomic(const cpp2::AddEdgesRequest& req) {
-    auto* processor = AddEdgesAtomicProcessor::instance(env_);
+    auto* processor = AddEdgesAtomicProcessor::instance(env_, &kAddEdgesAtomicCounters);
     RETURN_FUTURE(processor);
 }
 
--- a/src/kvstore/SnapshotManagerImpl.cpp
+++ b/src/kvstore/SnapshotManagerImpl.cpp
@@ -22,12 +22,16 @@ void SnapshotManagerImpl::accessAllRowsInSnapshot(GraphSpaceID spaceId,
     int64_t totalCount = 0;
 
     for (const auto& prefix : tables) {
-        accessTable(spaceId, partId, prefix, cb, data, totalCount, totalSize);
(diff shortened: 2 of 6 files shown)


bright-starry-sky (approved):
Nice, LGTM.

bright-starry-sky:
A question, how about "rvalue reference" for  asyncAppendBatch？WDYT? such as : 
```
virtual void asyncAppendBatch(GraphSpaceID spaceId, PartitionID partId, std::string&& batch, KVCallback cb) = 0;
folly::SemiFuture<kvstore::ResultCode> commitBatch(GraphSpaceID spaceId, PartitionID partId, std::string&& batch);
```

critical27:
> A question, how about "rvalue reference" for asyncAppendBatch？WDYT? such as :
> 
> ```
> virtual void asyncAppendBatch(GraphSpaceID spaceId, PartitionID partId, std::string&& batch, KVCallback cb) = 0;
> folly::SemiFuture<kvstore::ResultCode> commitBatch(GraphSpaceID spaceId, PartitionID partId, std::string&& batch);
> ```

I have thought about this, we could unify all interface in KVStore later, for now it seems all methods are using `T` or instead of `T&&`, such as asyncMultiPut.

bright-starry-sky (approved):
LGTM

src/storage/GraphStorageServiceHandler.cpp
```diff
@@ -89,14 +89,14 @@ GraphStorageServiceHandler::future_updateVertex(const cpp2::UpdateVertexRequest&
 // Edge section
 folly::Future<cpp2::ExecResponse>
 GraphStorageServiceHandler::future_addEdges(const cpp2::AddEdgesRequest& req) {
-    auto* processor = AddEdgesProcessor::instance(env_);
+    auto* processor = AddEdgesProcessor::instance(env_, &kAddEdgesCounters);
```
panda-sheep, line 92: No need to increase here, because the default value is used

src/storage/GraphStorageServiceHandler.cpp
```diff
@@ -89,14 +89,14 @@ GraphStorageServiceHandler::future_updateVertex(const cpp2::UpdateVertexRequest&
...
     RETURN_FUTURE(processor);
 }
 
 
 folly::Future<cpp2::ExecResponse>
 GraphStorageServiceHandler::future_deleteEdges(const cpp2::DeleteEdgesRequest& req) {
-    auto* processor = DeleteEdgesProcessor::instance(env_);
+    auto* processor = DeleteEdgesProcessor::instance(env_, &kDelEdgesCounters);
```
panda-sheep, line 99: ditto

src/storage/admin/RebuildEdgeIndexTask.cpp
```diff
@@ -116,6 +116,7 @@ kvstore::ResultCode RebuildEdgeIndexTask::buildIndexGlobal(GraphSpaceID space,
                     IndexKeyUtils::collectIndexValues(reader.get(), item->get_fields());
                 if (!valuesRet.ok()) {
                     LOG(WARNING) << "Collect index value failed";
+                    iter->next();
```
panda-sheep, line 119: Isn't it necessary here? 
Because the current index is wrong, other indexes are not necessarily wrong

src/storage/admin/RebuildTagIndexTask.cpp
```diff
@@ -103,6 +103,7 @@ kvstore::ResultCode RebuildTagIndexTask::buildIndexGlobal(GraphSpaceID space,
                     IndexKeyUtils::collectIndexValues(reader.get(), item->get_fields());
                 if (!valuesRet.ok()) {
                     LOG(WARNING) << "Collect index value failed";
+                    iter->next();
```
panda-sheep, line 106: ditto

src/storage/GraphStorageServiceHandler.cpp
critical27: Just keep it, explicit is better. It seems all processor has default value.

src/storage/admin/RebuildEdgeIndexTask.cpp
critical27: It will infinite loop if we go to here.

panda-sheep: Firstly, there may be multiple indexes, the current index is wrong, `next(),` the following index is not processed.
Secondly, the index is processed on line 132

critical27: Ah, sorry, I pushed the wrong commit, this has been fixed before. Thx.

bright-starry-sky (approved):

panda-sheep (approved):
LGTM

PR #346, opened 2021-02-04
Memory Lock, based on folly::ConcurrentHashMap

liuyu85cn:
Background:

Some storage operations in nebula has to be done in queue. 
for example, 
update one vertex(edge), currently we  wrap these operations in "AtomicOp".
performed by raft, heavy and slow.

But, nebula is built on Raft, 
which means all these things will happen only in "Leader"
we believe if we set proper memory locks in leader, then everything work fine. 

==================================================================

Just commit this PR to collect more request. 

Let's have a brain storm.

Any suggestion is welcomed.


src/storage/transaction/MemoryLockWrapper.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include "storage/transaction/MemoryLockCore.h"
+
+namespace nebula {
+namespace storage {
+
+template<class Key>
+class SimpleMemoryLock {
+public:
+    explicit SimpleMemoryLock(const Key& key) : SimpleMemoryLock(std::vector<Key>{key}) {}
```
critical27, line 17: Better rename to sth like `LockBatch` or else? `Lock lk1("1");` is not a lock.

src/storage/transaction/MemoryLockWrapper.h
```diff
@@ -0,0 +1,43 @@
...
+        locked_ = GlobalMemoryLock<Key>::inst().lockBatch(keys_);
+    }
+
+    ~SimpleMemoryLock() {
+        if (locked_) {
+            GlobalMemoryLock<Key>::inst().unlockBatch(keys_);
+        }
+    }
+
+    bool isLock() const noexcept {
```
critical27, line 29: `isLocked`?

critical27 (commented):
I think this is enough for our scenario. Good job

critical27:
Better to move to utils. @bright-starry-sky, plz have a look?

src/storage/transaction/MemoryLockCore.h
```diff
@@ -0,0 +1,82 @@
...
+        hashMap_.clear();
+    }
+
+protected:
+    folly::ConcurrentHashMap<Key, int> hashMap_;
+};
+
+
+template<typename Key>
+class GlobalMemoryLock {
```
critical27, line 70: Just MemoryLockCore<Key> is enough, no need to be singleton.

src/utils/MemoryLockWrapper.h
```diff
@@ -0,0 +1,45 @@
...
+namespace nebula {
+
+// RAII style to easily control the lock acquire / release
+template<class Key>
+class MemoryLockGuard {
+public:
+    MemoryLockGuard(MemoryLockCore<Key>* lock, const Key& key)
+        : MemoryLockGuard(lock, std::vector<Key>{key}) {}
+
+    explicit MemoryLockGuard(MemoryLockCore<Key>* lock, const std::vector<Key>& key)
```
bright-starry-sky, line 20: Great job. all are good for me.
A small suggestion, could we add a new constructor method as below?
`  explicit MemoryLockGuard(MemoryLockCore<Key>* lock, const std::set<Key>& key)`
A scene like this, Insert duplicate values in insert request, such as : 
`INSERT VERTEX t1(c1) VALUES "1":("1"), "1":("2"), 1:("3")`
the `key` should be {"1"},  not is {"1", "1", "1"}
Will be easier to use. WDYT?

src/utils/MemoryLockWrapper.h
```diff
@@ -0,0 +1,45 @@
...
+        : lock_(lock), keys_(key) {
+        locked_ = lock->lockBatch(keys_);
```
bright-starry-sky, line 22: Could return a pair<iter, bool> for lock->lockBatch(keys_)? We need to know which key is in conflict.

liuyu85cn: good catch.

src/utils/MemoryLockWrapper.h
liuyu85cn: I add a constructor, which has a bool parameter 'dedup'.

bright-starry-sky (approved):
LGTM. nice job.

critical27 (approved):
Awesome!

PR #348, opened 2021-02-05
adjust rocksdb compaction parameter

panda-sheep:
As title!

Please see the description below:
 ```
  // Compact the underlying storage for the key range [*begin,*end].
  // The actual compaction interval might be superset of [*begin, *end].
  // In particular, deleted and overwritten versions are discarded,
  // and the data is rearranged to reduce the cost of operations
  // needed to access the data.  This operation should typically only
  // be invoked by users who understand the underlying implementation.
  //
  // begin==nullptr is treated as a key before all keys in the database.
  // end==nullptr is treated as a key after all keys in the database.
  // Therefore the following call will compact the entire database:
  //    db->CompactRange(options, nullptr, nullptr);
  // Note that after the entire database is compacted, all data are pushed
  // down to the last level containing any data. If the total data size after
  // compaction is reduced, that level might not be appropriate for hosting all
  // the files. In this case, client could set options.change_level to true, to
  // move the files back to the minimum level capable of holding the data set
  // or a given level (specified by non-negative options.target_level).
  virtual Status CompactRange(const CompactRangeOptions& options,
                              ColumnFamilyHandle* column_family,
                              const Slice* begin, const Slice* end) = 0;
                             
  // CompactRangeOptions is used by CompactRange() call.
struct CompactRangeOptions {
  // If true, no other compaction will run at the same time as this
  // manual compaction
  bool exclusive_manual_compaction = true;
  // If true, compacted files will be moved to the minimum level capable
  // of holding the data or given level (specified non-negative target_level).
  bool change_level = false;
  // If change_level is true and target_level have non-negative value, compacted
  // files will be moved to target_level.
  int target_level = -1;
```

### CODE REVIEW

File: src/kvstore/RocksEngine.cpp [hunk 1]
```diff
@@ -363,6 +363,10 @@ ResultCode RocksEngine::setDBOption(const std::string& configKey, const std::str
 
 ResultCode RocksEngine::compact() {
     rocksdb::CompactRangeOptions options;
+    // If change_level is true and target_level is -1,
+    // compacted files will be moved to the minimum level capable of holding the data.
+    options.change_level = true;
+    options.target_level = -1;
```
critical27, line 369: You'd better add a default parameter, by default use `CompactRangeOptions()`

panda-sheep: ok!

critical27 (approved):

liuyu85cn (approved):

PR #349, opened 2021-02-07
Config the upgrade string index key length limit.

Shylock-Hg:
NG-450

### DIFF
--- a/src/meta/MetaVersionMan.cpp
+++ b/src/meta/MetaVersionMan.cpp
@@ -12,6 +12,7 @@
 
 DEFINE_bool(null_type, true, "set schema to support null type");
 DEFINE_bool(print_info, false, "enable to print the rewrite data");
+DEFINE_uint32(string_index_limit, 256, "string index key length limit");
 
 namespace nebula {
 namespace meta {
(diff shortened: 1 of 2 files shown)


critical27:
I think use 128 or 64 as default value is enough? @bright-starry-sky WDYT?

Shylock-Hg:
> I think use 128 or 64 as default value is enough? @bright-starry-sky WDYT?

Fixed.

src/meta/MetaVersionMan.cpp
```diff
@@ -12,6 +12,7 @@
 
 DEFINE_bool(null_type, true, "set schema to support null type");
 DEFINE_bool(print_info, false, "enable to print the rewrite data");
+DEFINE_uint32(string_index_limit, 64, "string index key length limit");
```
laura-ding, line 15: how about `index_length`, it's more simple and clear

Shylock-Hg: It's only used for string type.

jievince (approved):

critical27 (approved):

PR #351, opened 2021-02-08
Encode rank to make key string in order.

guojun85:


critical27:
I understand that rank is not sorted. But I'm afraid that if we change to big endian, all users's data is changed. And I believe there are many interfaces is affected, you can see check by the test cases by running `ctest`.

What is your reason to change this encode?

guojun85:
https://discuss.nebula-graph.com.cn/t/topic/2774/9

for this discuss, I want implement paging functionality to use "go from "1" over follow where follow._dst > "100000" and follow.rank > 3 |  limit 100000". but rank is not in ordered.

I have run unit testing, only one test case "rocks_engine_config_test" is failed which is not related to this change.

65/69 Test #65: storage_http_ingest_test .........   Passed    1.24 sec
      Start 66: scan_edge_test
66/69 Test #66: scan_edge_test ...................   Passed    2.24 sec
      Start 67: scan_vertex_test
67/69 Test #67: scan_vertex_test .................   Passed    2.15 sec
      Start 68: kv_test
68/69 Test #68: kv_test ..........................   Passed    1.99 sec
      Start 69: kv_client_test
69/69 Test #69: kv_client_test ...................   Passed    6.81 sec

99% tests passed, 1 tests failed out of 69

Label Time Summary:
codec             =   0.03 sec*proc (4 tests)
kvstore           =  95.98 sec*proc (7 tests)
kvstore/raftex    = 267.75 sec*proc (9 tests)
kvstore/wal       =   9.41 sec*proc (4 tests)
meta              = 185.12 sec*proc (16 tests)
storage           = 174.91 sec*proc (26 tests)
utils             =   0.02 sec*proc (3 tests)

Total Test time (real) = 733.25 sec
The following tests FAILED:
         25 - rocks_engine_config_test (Failed)

guojun85:
For keys in V2.0 have already been changed, and upgrade is need from V1.0. Why not further to make edge key in order?

guojun85:
ctest fail seems not because of this change.

src/utils/NebulaKeyUtils.cpp
```diff
@@ -51,7 +51,7 @@ std::string NebulaKeyUtils::edgeKey(size_t vIdLen,
        .append(srcId.data(), srcId.size())
        .append(vIdLen - srcId.size(), '\0')
        .append(reinterpret_cast<const char*>(&type), sizeof(EdgeType))
-       .append(reinterpret_cast<const char*>(&rank), sizeof(EdgeRanking))
+       .append(IndexKeyUtils::encodeInt64(rank))
```
panda-sheep, line 54: First of all, I think it is inappropriate to use the` IndexKeyUtils::encodeInt64` interface here

ps： If this is changed, a new interface is needed, right? Or this interface needs to be renamed

guojun85: How about use IndexKeyUtils::encodeValue?

critical27:
> For keys in V2.0 have already been changed, and upgrade is need from V1.0. Why not further to make edge key in order?

Yeah, it is the right time if we need to change key format (If we really need to do it). Let's talk in the forum.

src/utils/NebulaKeyUtils.cpp
critical27: I think just copy the IndexKeyUtils::encodeInt64 and add a new function like NebulaKeyUtils::encodeRank?

guojun85: OK, will change it.

src/utils/IndexKeyUtils.h
```diff
@@ -165,6 +165,22 @@ class IndexKeyUtils final {
         return val;
     }
 
+    static std::string encodeRank(EdgeRanking rank) {
```
critical27, line 168: Would you move it to `NebulaKeyUtils`? So `NebulaKeyUtils` don't need to depend on `IndexKeyUtils`

guojun85: but IndexKeyUtils also used encodeRank, if move to NebulaKeyUtils, IndexKeyUtils will need to depends on NebulaKeyUtils.

And encode and decode series of functions currently are in IndexKeyUtils, I think it's better to put them together in IndexKeyUtils.

critical27: `IndexKeyUtils` are used only for index encoding, I suppose both `IndexKeyUtils` and `NebulaKeyUtils` have a method `encodeRank` is clearer. If one depends on another, which will bring more compile time.

guojun85: OK, will change it according to your comments.

src/utils/IndexKeyUtils.h
```diff
@@ -165,6 +165,22 @@ class IndexKeyUtils final {
...
+        return raw;
+    }
+
+    static EdgeRanking decodeRank(const folly::StringPiece& raw) {
+        auto val = *reinterpret_cast<const int64_t*>(raw.data());
+        val = folly::Endian::big(val);
+        val ^= folly::to<int64_t>(1) << 63;
+        return val;
+    }
+
```
bright-starry-sky, line 183: Please add test cases for bound value

guojun85: OK

src/utils/IndexKeyUtils.h
panda-sheep: `encodeRank `can directly call `encodeInt64`

src/utils/IndexKeyUtils.h
```diff
@@ -165,6 +165,22 @@ class IndexKeyUtils final {
...
+    static std::string encodeRank(EdgeRanking rank) {
+        rank ^= folly::to<int64_t>(1) << 63;
+        auto val = folly::Endian::big(rank);
+        std::string raw;
+        raw.reserve(sizeof(int64_t));
+        raw.append(reinterpret_cast<const char*>(&val), sizeof(int64_t));
+        return raw;
+    }
+
+    static EdgeRanking decodeRank(const folly::StringPiece& raw) {
```
panda-sheep, line 177: ditto

guojun85: OK

panda-sheep (approved):
Good job，thanks for your contribution！

critical27 (approved):

PR #352, opened 2021-02-08
 change toBool

czpmango:
The interface `Value::toBool` in nebula-common has been modified, so it needs to be modified accordingly. 
But I don’t understand why `toBool` is needed here.

Depends on [https://github.com/vesoft-inc/nebula-common/pull/412](https://github.com/vesoft-inc/nebula-common/pull/412)

### DIFF
--- a/src/storage/exec/FilterNode.h
+++ b/src/storage/exec/FilterNode.h
@@ -68,7 +68,7 @@ private:
             auto result = filterExp_->eval(*expCtx_);
             // NULL is always false
             auto ret = result.toBool();
-            if (ret.second && ret.first) {
+            if (ret.isBool() && ret.getBool()) {
                 return true;
             }
             return false;


src/storage/exec/FilterNode.h
```diff
@@ -68,7 +68,7 @@ class FilterNode : public IterateNode<T> {
             auto result = filterExp_->eval(*expCtx_);
             // NULL is always false
             auto ret = result.toBool();
```
czpmango, line 70: Is line 70 necessary？

panda-sheep: Yeah, need

panda-sheep:
please hold this pr, the toBool interface does not meet the scene here

czpmango:
> please hold this pr, the toBool interface does not meet the scene here

The related PR in nebula-common has been marked as `DO NOT MERGE`.
The `toBool` interface will not be changed until the storage side is completed.

critical27 (dismissed):

panda-sheep (approved):
please handle` https://github.com/vesoft-inc/nebula-graph/issues/762` this situation. thx

src/storage/exec/FilterNode.h
```diff
@@ -68,7 +68,7 @@ class FilterNode : public IterateNode<T> {
...
-            if (ret.second && ret.first) {
```
panda-sheep, line 71: Because the where/when conditions of this go/update/upsert are all handled here

If you only support the go where condition, the field must be of type bool or string.
Fields that do not support the when condition of update/upsert must be of bool type or string type. The update/upsert when condition will be filtered out here instead of a syntax error.

so, please handle the when and where conditions of go. update upsert uniformly.

panda-sheep (changes requested):

src/storage/exec/FilterNode.h
czpmango: > Because the `where/when` conditions of this `go/update/upsert` are all handled here
> 
> If you only support the `go where` condition, the field must be of type` bool` or `string`.
> Fields that must be of bool type or string type do not handle the `when `condition of `update/upsert` .
> The `update/upsert` `when `condition will be filtered out here instead of a syntax error.
> Causes an error in the result of the `update/upsert` statement
> 
> so, please handle the `when and where` conditions of `go update upsert` uniformly.
> @laura-ding @czpmango

ACK.

czpmango: Fixed in [https://github.com/vesoft-inc/nebula-graph/pull/772](https://github.com/vesoft-inc/nebula-graph/pull/772) 
and [https://github.com/vesoft-inc/nebula-graph/pull/767](https://github.com/vesoft-inc/nebula-graph/pull/767)

yixinglu (approved):

critical27 (approved):

CPWstatic (approved):

panda-sheep (approved):
LGTM

PR #353, opened 2021-02-08
Unify all getActiveHosts using heartbeat_interval_secs

critical27:
Previously we use `expired_threshold_sec` or `heartbeat_interval_secs` in `ActiveHostsMan::getActiveHosts`. It will bring some problems, for example: show hosts use `expired_threshold_sec`, and as for default value, it takes 10 minutes to see storage is OFFLINE after stopped.

For simplicity, we just all use `heartbeat_interval_secs` is enough, by default we will regard any graph/storage is online when meta received heartbeat within `2 * heartbeat_interval_secs`.


src/meta/ActiveHostsMan.cpp
```diff
@@ -49,7 +48,7 @@ std::vector<HostAddr> ActiveHostsMan::getActiveHosts(kvstore::KVStore* kv,
         FLOG_ERROR("getActiveHosts failed(%d)", static_cast<int>(ret));
         return hosts;
     }
-    int64_t threshold = (expiredTTL == 0 ? FLAGS_expired_threshold_sec : expiredTTL) * 1000;
+    int64_t threshold = (expiredTTL == 0 ? FLAGS_heartbeat_interval_secs * 2 : expiredTTL) * 1000;
```
laura-ding, line 51: I think it would be better to increase the time, like 5, the 2 is so small. Or support configurable.

critical27: Actually, we have been use `FLAGS_heartbeat_interval_secs * 2` in some interfaces for a long time, such as `show hosts`. I believe it is enough.

panda-sheep (approved):
Good job，make the code cleaner！

bright-starry-sky (approved):
LGTM

PR #355, opened 2021-02-18
Improved performance with index for insert, delete and update

bright-starry-sky:
Total rows 14349843 | AVG(Rows/s) | CPU | Time spent | performance degradation
-- | -- | -- | -- | --
original | 18612.82/s | 897.3 | 771.47s | 53384.70/s 18622.70/s
improved | 59050.95/s | 2750 | 243.51s | 117893.05/s 59459.68/s


src/storage/CommonUtils.h
```diff
@@ -59,6 +60,10 @@ using VertexCache = ConcurrentLRUCache<std::pair<VertexID, TagID>, std::string>;
 using IndexKey    = std::tuple<GraphSpaceID, PartitionID>;
 using IndexGuard  = folly::ConcurrentHashMap<IndexKey, IndexState>;
 
+using VMLI = std::tuple<GraphSpaceID, PartitionID, TagID, VertexID>;
```
critical27, line 63: Have you try for string key? I don't know how much difference with tuple and string in ConcurrentHashMap benchmark.

critical27 (commented):
Good job!

src/storage/exec/UpdateNode.h
```diff
@@ -167,57 +167,64 @@ class UpdateTagNode : public UpdateNode<VertexID> {
...
+                                                       partId, tagId_, vId)};
+        nebula::MemoryLockGuard<VMLI> lg(planContext_->env_->verticesML_.get(), dummyLock);
+        if (!lg) {
+            auto conflict = lg.conflictKey();
+            LOG(ERROR) << "vertex conflict "
+                       << std::get<0>(conflict) << ":"
+                       << std::get<1>(conflict) << ":"
+                       << std::get<2>(conflict) << ":"
+                       << std::get<3>(conflict);
+            return kvstore::ResultCode::ERR_CONSENSUS_ERROR;
```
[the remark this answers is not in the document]
bright-starry-sky: > Better return another error code? Storage client could retry if we add retry later.

I think here is no need to retry always. we need to let the user know where the conflict is, user retries at the business layer. 
Also, if we need to retry, the best way is just skip this conflict row and report warning, not need return error code. WDYT?

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -100,22 +110,163 @@ void AddEdgesProcessor::process(const cpp2::AddEdgesRequest& req) {
...
+            batchHolder->put(std::move(key), std::move(retEnc.value()));
+            dummyLock.emplace_back(std::make_tuple(spaceId_,
+                                                   partId,
+                                                   edgeKey.src.getStr(),
+                                                   edgeKey.edge_type,
+                                                   edgeKey.ranking,
+                                                   edgeKey.dst.getStr()));
+        }
+        auto batch = encodeBatchValue(std::move(batchHolder)->getBatch());
+        if (batch.empty()) {
```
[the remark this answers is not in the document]
bright-starry-sky: > batch should not be empty unless there are no edge key in a part? If so, we don't need to call `handleAsync`.

In fact, the batch cannot be empty，this is just a fault-tolerant. maybe here can be add a DCHECK, WDYT?

src/storage/CommonUtils.h
bright-starry-sky: > Have you try for string key? I don't know how much difference with tuple and string in ConcurrentHashMap benchmark.

Your means, instead it to tuple<spaceId, string_key>?
This has been tried before in a draft version and has not improved performance. The current implementation has two advantages, 1) easier access to conflict item (Vid, src, dst, edgeType, rank) for error report. 2) more intuitive, edge version(1 byte) are easier to control.

src/storage/mutate/AddEdgesProcessor.cpp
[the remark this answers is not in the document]
critical27: DCHECK is fine.

src/storage/exec/UpdateNode.h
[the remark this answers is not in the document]
critical27: Why do you think we should not retry when conflicts? 
I mean if we add retry later, when conflict, storage client should take some special handling just like leader change?

src/storage/CommonUtils.h
critical27: Ok.

src/storage/exec/UpdateNode.h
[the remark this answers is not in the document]
bright-starry-sky: > Why do you think we should not retry when conflicts?
> I mean if we add retry later, when conflict, storage client should take some special handling just like leader change?
Because we don't know what the final data is if there's a retry
for example : insert vertex , the vertex id is 1
retry : 
first time
thread 1 : insert vertex 1 (conflicted)
thread 2 : insert vertex 1 (conflicted)
second time
thread 1 : insert vertex 1 (success | conflicted then retry)
thread 2 : insert vertex 1 (success  | conflicted then retry)
if conflicted , then retry ... it was successful in the end
the final data is thread 1's data, right? Or is it thread 2?

If retry is supported, there is a better way：
first time
thread 1 : insert vertex 1 (conflicted, skip the row in mem lock and report warning, then write data immediate)
thread 2 : insert vertex 1 (conflicted, skip the row in mem lock and report warning, then write data immediate)
it was successful same with retry.

src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -108,132 +117,162 @@ void AddVerticesProcessor::process(const cpp2::AddVerticesRequest& req) {
...
+                           << std::get<2>(conflict) << ":"
+                           << std::get<3>(conflict);
+                pushResultCode(cpp2::ErrorCode::E_CONSENSUS_ERROR, partId);
+                onFinished();
+                return;
             }
+            auto callback = [partId, this](kvstore::ResultCode code) {
+                handleAsync(spaceId_, partId, code);
+            };
+            env_->kvstore_->asyncAppendBatch(spaceId_, partId, std::move(batch), callback);
```
critical27, line 273: It seems `MemoryLockGuard` life cycle is short than `asyncAppendBatch`, would something go wrong?

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -100,22 +110,163 @@ void AddEdgesProcessor::process(const cpp2::AddEdgesRequest& req) {
...
+    const auto& propNames = req.get_prop_names();
+    for (auto& part : partEdges) {
+        IndexCountWrapper wrapper(env_);
+        std::unique_ptr<kvstore::BatchHolder> batchHolder =
+        std::make_unique<kvstore::BatchHolder>();
+        std::vector<EMLI> dummyLock;
+        auto partId = part.first;
+        const auto& newEdges = part.second;
+        for (auto& newEdge : newEdges) {
+            auto edgeKey = newEdge.key;
```
critical27, line 128: `dummyLock` could reserve here.

src/storage/mutate/AddVerticesProcessor.cpp
bright-starry-sky: > It seems `MemoryLockGuard` life cycle is shorter than `asyncAppendBatch`, would something go wrong?
> I suppose unlock should happen in callback.
Good point!

src/storage/exec/UpdateNode.h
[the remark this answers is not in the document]
critical27: Yeah, retry logic make sense. Anyway, the result code need to change, `ERR_CONSENSUS_ERROR` is used in other case.

src/storage/mutate/AddVerticesProcessor.cpp
bright-starry-sky: > It seems `MemoryLockGuard` life cycle is shorter than `asyncAppendBatch`, would something go wrong?
> I suppose unlock should happen in callback.

good point

src/storage/exec/UpdateNode.h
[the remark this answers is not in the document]
bright-starry-sky: > Yeah, retry logic make sense. Anyway, the result code need to change, `ERR_CONSENSUS_ERROR` is used in other case.

make sense, what's about `ERR_CONFLICT_DATA`?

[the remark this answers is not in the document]
critical27: 👌

src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -108,132 +117,162 @@ void AddVerticesProcessor::process(const cpp2::AddVerticesRequest& req) {
...
+                    VLOG(3) << "Evict cache for vId " << vid
+                            << ", tagId " << tagId;
+                }
+            }
+        }
+        auto batch = encodeBatchValue(std::move(batchHolder)->getBatch());
+        if (batch.empty()) {
+            handleAsync(spaceId_, partId, kvstore::ResultCode::SUCCEEDED);
+        } else {
+            nebula::MemoryLockGuard<VMLI> lg(env_->verticesML_.get(), dummyLock, true);
```
critical27, line 258: `dummyLock` could move as well

bright-starry-sky:
Addressed comments.
depends on common PR 417 since error code changed.
https://github.com/vesoft-inc/nebula-common/pull/417

src/utils/MemoryLockWrapper.h
```diff
@@ -30,6 +30,10 @@ class MemoryLockGuard {
         }
     }
 
+    MemoryLockGuard(MemoryLockGuard&& lg)
+        : lock_(lg.lock_), keys_(std::move(lg.keys_)), locked_(lg.locked_) {}
+
+
```
critical27, line 36: Add the `operator=(MemoryLockGuard&&)` well, `nocept` is necessary as well. Forbid the copy ctor and operator=.

bright-starry-sky: > Add the `operator=(MemoryLockGuard&&)` well, `noexcept` is necessary as well. Forbid the copy ctor and operator=.

Assent!

bright-starry-sky:
fixed error of constructor method.

src/utils/MemoryLockWrapper.h
```diff
@@ -30,7 +30,24 @@ class MemoryLockGuard {
...
+        if (this != &lg) {
+            lock_ = lg.lock_;
+            keys_ = std::move(lg.keys_);
+            locked_ = lg.locked_;
+        }
+        return *this;
+    }
+
     ~MemoryLockGuard() {
+        LOG(INFO) << "begin ~MemoryLockGuard";
```
critical27, line 50: For debug? Maybe lower the log level.

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -100,22 +110,162 @@ void AddEdgesProcessor::process(const cpp2::AddEdgesRequest& req) {
...
+                                                   edgeKey.edge_type,
+                                                   edgeKey.ranking,
+                                                   edgeKey.dst.getStr()));
+        }
+        auto batch = encodeBatchValue(std::move(batchHolder)->getBatch());
+        DCHECK(!batch.empty());
+        nebula::MemoryLockGuard<EMLI> lg(env_->edgesML_.get(), std::move(dummyLock), true);
+        if (!lg) {
+            auto conflict = lg.conflictKey();
+            LOG(ERROR) << "edge conflict "
```
critical27, line 249: Do we need to lower this log level? WARNING or VLOG?

critical27 (commented):
LGTM now, @vesoft-inc/storage-reviewers PTAL

src/storage/mutate/AddEdgesProcessor.cpp
bright-starry-sky: > Do we need to lower this log level? WARNING or VLOG?

keep it, let me change them after retry supported.

bright-starry-sky:
Removed debug info.

src/storage/mutate/AddEdgesProcessor.cpp
critical27: 👌

critical27 (approved):

critical27:
We'd better have a test on the memory lock key type, I remember that hash function for a tuple is really slow, I don't know if concurrent hashmap will be a bottleneck.

src/storage/exec/UpdateNode.h
```diff
@@ -167,57 +167,65 @@ class UpdateTagNode : public UpdateNode<VertexID> {
...
+            LOG(ERROR) << "vertex conflict "
+                       << std::get<0>(conflict) << ":"
+                       << std::get<1>(conflict) << ":"
+                       << std::get<2>(conflict) << ":"
+                       << std::get<3>(conflict);
+            return kvstore::ResultCode::ERR_DATA_CONFLICT_ERROR;
+        }
+
+        folly::Baton<true, std::atomic> baton;
+        auto callback = [&ret, &baton] (kvstore::ResultCode code) {
```
panda-sheep, line 223: Please put `lg` in the callback function to avoid inconsistent life cycle.

src/storage/exec/UpdateNode.h
```diff
@@ -520,8 +516,32 @@ class UpdateEdgeNode : public UpdateNode<cpp2::EdgeKey> {
...
+                           << std::get<0>(conflict) << ":"
+                           << std::get<1>(conflict) << ":"
+                           << std::get<2>(conflict) << ":"
+                           << std::get<3>(conflict) << ":"
+                           << std::get<4>(conflict) << ":"
+                           << std::get<5>(conflict);
+                return kvstore::ResultCode::ERR_DATA_CONFLICT_ERROR;
+            }
+            folly::Baton<true, std::atomic> baton;
+            auto callback = [&ret, &baton] (kvstore::ResultCode code) {
```
panda-sheep, line 539: ditto

panda-sheep (commented):
Good job, well done!
Please refer to the following notes.

src/storage/exec/UpdateNode.h
bright-starry-sky: > Please put `lg` in the callback function to avoid inconsistent life cycle.

Aha, don't need put 'lg' to callback at here. `folly::Baton` will control the lifecycle.

bright-starry-sky:
> We'd better have a test on the memory lock key type, I remember that hash function for a tuple is really slow, I don't know if concurrent hashmap will be a bottleneck. Or we check how much does hashmap takes in perf at first.

Well thought out. seems that the current logic is acceptable. I've tested via #359 .

src/storage/exec/UpdateNode.h
panda-sheep: yeah. `baton.wait()`  has waited.

panda-sheep (approved):

czpmango:
Did this PR pass the CI test of nebula-graph? @bright-starry-sky

PR #358, opened 2021-02-23
PlainTable support

critical27:
To support rocksdb PlainTable and nebula work in memory
1. Dir of rocksdb wal and nebula wal should be configurable
2. Backup rocksdb periodically, and try to recover from latest backup
3. Fix the 2.0 prefix bloom filter

### DIFF
--- a/src/kvstore/RocksEngine.cpp
+++ b/src/kvstore/RocksEngine.cpp
@@ -80,11 +80,12 @@ public:
 RocksEngine::RocksEngine(GraphSpaceID spaceId,
                          int32_t vIdLen,
                          const std::string& dataPath,
+                         const std::string& walPath,
                          std::shared_ptr<rocksdb::MergeOperator> mergeOp,
                          std::shared_ptr<rocksdb::CompactionFilterFactory> cfFactory,
                          bool readonly)
     : KVEngine(spaceId)
-    , spaceId_(spaceId),
+    , spaceId_(spaceId)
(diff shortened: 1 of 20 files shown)


src/kvstore/RocksEngine.cpp
```diff
@@ -114,6 +125,8 @@ RocksEngine::RocksEngine(GraphSpaceID spaceId,
     db_.reset(db);
     partsNum_ = allParts().size();
     LOG(INFO) << "open rocksdb on " << path;
+
+    backup();
```
bright-starry-sky, line 129: A little confused, seems like a run for nothing, could give some comments?

critical27: Long story, I can't remember very clear.
1. If we don't backup here, we write some data, and storage is stopped before we do a valid backup.
2. Mock machine reboot, so rocksdb data in memory is lost. (rocksdb wal is persisted in disk, so we could recover from it in theory)
3. Restart the storage, rocks db will be failed open, log is `While creating a new Db, wal_dir contains existing log file: ...`.

I believe the problem is that if we backup here, it is not nothing here. Something like `CURRENT` and `MANIFEST` will be backuped. In that way, rocksdb will think the wal is legal.

bright-starry-sky: I see. It's a very complicated story.

bright-starry-sky (dismissed):
Good job. Well done.

bright-starry-sky (approved):
LGTM

panda-sheep (approved):
Woohoo，Great job！
I learned a lot from it.👍

PR #338, opened 2021-01-28
make toss U-turn

liuyu85cn:
change the chain direction of toss from

out-edge=> in-edge => out-edge , to
in-edge => out-edge => in-edge.

=====================================================
please ignore the commented CMake things, those just for fast compile, will restore at last.

PR #360, opened 2021-02-24
Filter edge lock when getting edge

panda-sheep:
Filter edge lock when getting edge.
Because edge lock in toss and edge are the same except for the last byte.

### DIFF
--- a/src/storage/admin/RebuildTagIndexTask.cpp
+++ b/src/storage/admin/RebuildTagIndexTask.cpp
@@ -72,10 +72,6 @@ kvstore::ResultCode RebuildTagIndexTask::buildIndexGlobal(GraphSpaceID space,
 
         auto key = iter->key();
         auto val = iter->val();
-        if (!NebulaKeyUtils::isVertex(vidSize, key)) {
-            iter->next();
-            continue;
-        }
 
         auto tagID = NebulaKeyUtils::getTagId(vidSize, key);
 
--- a/src/storage/admin/StatisTask.cpp
+++ b/src/storage/admin/StatisTask.cpp
@@ -131,17 +131,15 @@ StatisTask::genSubTask(GraphSpaceID spaceId,
         edgetypeEdges[edge.first] = 0;
     }
 
-    TagID                                 lastTagId = 0;
     VertexID                              lastVertexId = "";
-    // Only statis valid vetex data
+    // Only statis valid vetex data, no multi version
     // For example
-    // Vid  tagId  version
-    // 1     1     2
-    // 1     1     1
-    // 2     2     2   (invalid data)
-    // 2     2     1   (invalid data)
-    // 2     3     1
-    // 3     1     2
+    // Vid  tagId
+    // 1     1
+    // 2     2  (invalid data)
+    // 2     3
+    // 2     5
+    // 3     1
     while (vertexIter && vertexIter->valid()) {
         auto key = vertexIter->key();
         auto vId = NebulaKeyUtils::getVertexId(vIdLen, key).str();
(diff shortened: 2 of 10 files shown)


src/storage/admin/RebuildTagIndexTask.cpp
```diff
@@ -72,10 +72,6 @@ kvstore::ResultCode RebuildTagIndexTask::buildIndexGlobal(GraphSpaceID space,
 
         auto key = iter->key();
         auto val = iter->val();
-        if (!NebulaKeyUtils::isVertex(vidSize, key)) {
```
critical27, line 75: Remove from `RebuildEdgeIndexTask.cpp` as well

src/storage/admin/StatisTask.cpp
```diff
@@ -185,6 +185,13 @@ StatisTask::genSubTask(GraphSpaceID spaceId,
...
     while (edgeIter && edgeIter->valid()) {
         auto key = edgeIter->key();
+
+        // Because edge lock in toss and edge are the same except for the last byte
+        if (!NebulaKeyUtils::isEdge(vIdLen, key)) {
+            edgeIter->next();
+            continue;
+        }
+
         auto edgeType = NebulaKeyUtils::getEdgeType(vIdLen, key);
```
critical27, line 195: You could remove the useless multi version check in count *

src/storage/query/ScanVertexProcessor.cpp
```diff
@@ -63,9 +63,6 @@ void ScanVertexProcessor::doProcess(const cpp2::ScanVertexRequest& req) {
     RowReaderWrapper reader;
     for (int64_t rowCount = 0; iter->valid() && rowCount < rowLimit; iter->next()) {
         auto key = iter->key();
-        if (!NebulaKeyUtils::isVertex(spaceVidLen_, key)) {
```
critical27, line 66: Remove it in ScanEdgeProcessor as well

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -392,8 +392,12 @@ AddEdgesProcessor::findOldValue(PartitionID partId, const folly::StringPiece& ra
                    << ", spaceId " << spaceId_;
         return folly::none;
     }
-    if (iter && iter->valid()) {
-        return iter->val().str();
+
+    while (iter && iter->valid()) {
```
critical27, line 396: No need to modify here. #338 will turn the direction. As for out edge, there won't be any lock.

critical27:
Good job, check other place if they call useless isVertex or isEdge

src/storage/admin/RebuildTagIndexTask.cpp
panda-sheep: Good catch!

src/storage/mutate/AddEdgesProcessor.cpp
panda-sheep: Yeah, 
btw Before merging #338 , please temporarily stop the use of toss

src/storage/query/ScanVertexProcessor.cpp
panda-sheep: ScanEdgeProcessor  can handle in-edge， so `isEdge  `needs.

src/storage/admin/StatisTask.cpp
panda-sheep: Good point！

src/storage/admin/StatisTask.cpp
```diff
@@ -131,17 +131,15 @@ StatisTask::genSubTask(GraphSpaceID spaceId,
...
-    // Vid  tagId  version
-    // 1     1     2
-    // 1     1     1
-    // 2     2     2   (invalid data)
-    // 2     2     1   (invalid data)
-    // 2     3     1
-    // 3     1     2
+    // Vid  tagId
+    // 1     1
+    // 2     2  (invalid data)
```
critical27, line 139: Tag 2 is dropped? why invalid, better add some comments.

critical27 (commented):
LGTM now

src/storage/admin/StatisTask.cpp
panda-sheep: Has added comments.
invalid tag data, for example, tag data without tag schema.  create tag, insert tag data, then drop tag, so this tag data is invalid.

critical27 (approved):
Good job

liuyu85cn (approved):
LGTM

PR #361, opened 2021-02-24
Fix the tag write logical with index.

Shylock-Hg:
See https://github.com/vesoft-inc/nebula-graph/runs/1967100738?check_suite_focus=true


CPWstatic (approved):

Aiee (approved):

critical27:
Good catch

bright-starry-sky:
Good catch , My fault. thanks for your fixed.

PR #364, opened 2021-02-25
update iterate edges benchmark

critical27:
Since we update storage data from 1.x to 2.x, actually we don't need `RowReaderWrapper` or `RowReaderV1` anymore. Based on the benchmark, performance impact is very small, so not in a rush, we could do it later.

When 10000 edges to iterate:
* RowReaderWrapper reset with vector schemas takes 10550us 
* RowReaderV2 reset with vector schemas takes 10182us

When 100 edges to iterate:
* RowReaderWrapper reset with vector schemas takes 87us 
* RowReaderV2 reset with vector schemas takes 86us


bright-starry-sky (approved):
LGTM

panda-sheep (approved):
LGTM

PR #367, opened 2021-03-01
Support leader balance times

darionyaphet:


critical27:
Why?

darionyaphet:
> Why?

Infinite loops may cause problems that cannot exit

PR #370, opened 2021-03-03
forbid SO_REUSEPORT

critical27:
In 2.0, storage will use at least 4 ports. If send rpc to wrong port, weird things will happen. For example, no matter what query it is, we always get an error like `Storage Error: part: 6, error: E_RPC_FAILURE(-3).`

We just remove `SO_REUSEPORT`.

See https://github.com/vesoft-inc/nebula/pull/2094.


darionyaphet (approved):

panda-sheep (approved):
Well done！

PR #371, opened 2021-03-03
Unfreeze the limit of SST part-file's size

darionyaphet:
When the amount of data is very small or there is a serious skew in the data set.  

Maybe SST file does not exist when generator create it. So I lifted that restriction.


jude-zhu (approved):

critical27 (approved):
👍

PR #339, opened 2021-01-28
Balance data with job manager

darionyaphet:

PR #373, opened 2021-03-04
remove shuffle when balance data

critical27:
For example is in hosts [a, b, c], want move from c to d, then b to c. However, b to c is invoked at first.
Expected: [abc] -> [abd] -> [acd]
Actual: [abc] -> [ac] -> [ad]

Just remove the shuffle, the relative order of given part can't be changed during balance.

image
image


darionyaphet:
yeah I have remove it in #339

liuyu85cn (approved):

bright-starry-sky (approved):

PR #376, opened 2021-03-05
fix early onFinished

critical27:
See https://github.com/vesoft-inc/nebula-storage/runs/2036566131?check_suite_focus=true.

1. fix mem leak
2. replace Optional<string> with ErrorOr<ResultCode, string>, so we could report leader change when read.

We need to unify kvstore and thrift ErrorCode later, it is really a disaster.
Please review carefully, or tell me if there are any better ideas.

### DIFF
--- a/src/storage/BaseProcessor.h
+++ b/src/storage/BaseProcessor.h
@@ -102,6 +102,10 @@ protected:
                      PartitionID partId,
                      kvstore::ResultCode code);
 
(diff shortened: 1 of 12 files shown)


bright-starry-sky:
I'm in favor of this change, generally good to me of current changes. maybe we can talk about a better solution later.

bright-starry-sky (approved):

liuyu85cn (approved):
LGTM

PR #378, opened 2021-03-08
Set heartbeat log level to debug level

darionyaphet:
avoid print out lot of logs which looks like:

```
I0307 22:08:33.110009 48848 HBProcessor.cpp:41] Receive heartbeat from [0.0.0.0:13699], role = GRAPH
I0307 22:08:33.289899 48848 HBProcessor.cpp:41] Receive heartbeat from [192.168.8.211:55600], role = STORAGE
I0307 22:08:33.671300 48848 HBProcessor.cpp:41] Receive heartbeat from [192.168.8.211:55700], role = STORAGE
I0307 22:08:33.889750 48848 HBProcessor.cpp:41] Receive heartbeat from [192.168.8.211:55500], role = STORAGE
I0307 22:08:36.113718 48848 HBProcessor.cpp:41] Receive heartbeat from [0.0.0.0:13699], role = GRAPH
```

### DIFF
--- a/src/meta/processors/admin/HBProcessor.cpp
+++ b/src/meta/processors/admin/HBProcessor.cpp
@@ -38,8 +38,8 @@ void HBProcessor::process(const cpp2::HBReq& req) {
         return;
     }
 
-    LOG(INFO) << "Receive heartbeat from " << host
-              << ", role = " << meta::cpp2::_HostRole_VALUES_TO_NAMES.at(req.get_role());
+    VLOG(3) << "Receive heartbeat from " << host
+            << ", role = " << meta::cpp2::_HostRole_VALUES_TO_NAMES.at(req.get_role());
     auto ret = kvstore::ResultCode::SUCCEEDED;
     if (req.get_role() == cpp2::HostRole::STORAGE) {
         ClusterID peerCluserId = req.get_cluster_id();


liuyu85cn:
how about LOG_EVERY_N(INFO, 100)?

liuyu85cn (dismissed):

darionyaphet:
> how about LOG_EVERY_N(INFO, 100)?

As a long running service, although 1% has a lot of logs

bright-starry-sky (approved):
LGTM

liuyu85cn (approved):
LGTM

critical27:
Meta is ok? By default only 1 log every 10 seconds by each host.

darionyaphet:
> Meta is ok? By default only 1 log every 10 seconds by each host.

Sure Here is three storage instance and one graph instance report the heartbeat.

That's not the point. The heartbeat log level should be debug level

critical27:
Maybe VLOG(1) is enough, this log is every important when debug, WDYT?

darionyaphet:
> Maybe VLOG(1) is enough, this log is every important when debug, WDYT?

It's not an important log so update to `VLOG(3)`. IMO there is no difference between `V1` and `V3`

darionyaphet:
> IMO, this log is important, for example, how can you tell which meta is leader if lower the log level, maybe we could discussed later

If necessary, you can provide a command to query the leader of the metadata service. Is it a good idea to judge from the log?

critical27:
> > IMO, this log is important, for example, how can you tell which meta is leader if lower the log level, maybe we could discussed later
> 
> If necessary, you can provide a command to query the leader of the metadata service. Is it a good idea to judge from the log?

Ok, we could do it later, I think it is necessary.

PR #379, opened 2021-03-09
Support negative correlativity

darionyaphet:
dependent on: https://github.com/vesoft-inc/nebula-common/pull/427/


src/storage/admin/StatisTask.cpp
```diff
@@ -176,26 +177,35 @@ StatisTask::genSubTask(GraphSpaceID spaceId,
         auto edgeType = NebulaKeyUtils::getEdgeType(vIdLen, key);
         // Because edge lock in toss and edge are the same except for the last byte.
         // But only the in-edge has a lock.
-        if (edgeType < 0 || edgetypeEdges.find(edgeType) == edgetypeEdges.end()) {
```
panda-sheep, line 179: If edgeType is less than 0, no find search is performed. Return directly

panda-sheep: Hmm, the case where edgeType is less than 0 is used below

panda-sheep (dismissed):

panda-sheep (dismissed):

liuyu85cn (dismissed):

panda-sheep (approved):
Good

liuyu85cn (approved):

PR #383, opened 2021-03-10
Added timestamp to data val

bright-starry-sky:


src/codec/RowWriterV2.cpp
```diff
@@ -916,7 +916,10 @@ std::string RowWriterV2::processOutOfSpace() noexcept {
 }
 
 
-WriteResult RowWriterV2::finish() noexcept {
+WriteResult RowWriterV2::finish(int64_t ts) noexcept {
```
critical27, line 919: Why not use the actual timestamp? We don't need to modify it in later version.

src/codec/RowWriterV2.cpp
```diff
@@ -916,7 +916,10 @@ std::string RowWriterV2::processOutOfSpace() noexcept {
...
+    // The timestamp will be saved to the tail of buf_
+    buf_.append(reinterpret_cast<char*>(&ts), sizeof(int64_t));
```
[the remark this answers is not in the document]
bright-starry-sky: Good catch. I omitted something here. fixed.

src/codec/RowWriterV2.cpp
bright-starry-sky: Good idea.

bright-starry-sky:
Addressed comments.

src/codec/RowWriterV2.cpp
```diff
@@ -911,12 +912,19 @@ std::string RowWriterV2::processOutOfSpace() noexcept {
                reinterpret_cast<void*>(&strLen),
                sizeof(int32_t));
     }
-
+    // reset timestamp.
+    auto ts = time::WallClock::fastNowInMicroSec();
+    temp.append(reinterpret_cast<char*>(&ts), sizeof(int64_t));
```
critical27, line 917: We could set ts only once before line 940? Otherwise, we will set once here once in line 926.

src/codec/RowWriterV2.cpp
```diff
@@ -911,12 +912,19 @@ std::string RowWriterV2::processOutOfSpace() noexcept {
...
     return temp;
 }
 
 
 WriteResult RowWriterV2::finish() noexcept {
+    CHECK(!finished_) << "You have called finish()";
```
critical27, line 923: 👍

src/codec/RowWriterV2.cpp
bright-starry-sky: Good suggestion. and fixed the bug for `checkUnsetFields`

panda-sheep (approved):
Good job

critical27 (approved):

PR #388, opened 2021-03-15
Fix the fatal of deref nullptr.

Shylock-Hg:
IN https://github.com/vesoft-inc/nebula-common/pull/431, in some cases the client will return as nullptr.


Shylock-Hg:
Replace by https://github.com/vesoft-inc/nebula-common/pull/452

PR #391, opened 2021-03-16
fix bug in balance data

critical27:
1. Fix index is missing when balance data. Em, big bug since 1.0 ... It has been test in chaos now.
2. Fix a bug in `AtomicLogBuffer`, there is a memory usage issue in it, need to be fixed ASAP.

### DIFF
--- a/src/utils/NebulaKeyUtils.cpp
+++ b/src/utils/NebulaKeyUtils.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "utils/NebulaKeyUtils.h"
+#include "utils/IndexKeyUtils.h"
 
 namespace nebula {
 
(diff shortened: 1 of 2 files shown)


bright-starry-sky (approved):
Good job

src/utils/NebulaKeyUtils.cpp
```diff
@@ -193,6 +194,9 @@ std::vector<std::string> NebulaKeyUtils::snapshotPrefix(PartitionID partId) {
     } else {
         result.emplace_back(vertexPrefix(partId));
         result.emplace_back(edgePrefix(partId));
+        result.emplace_back(IndexKeyUtils::indexPrefix(partId));
```
darionyaphet, line 197: It will affect balance?

critical27: We didn't move index data :bug:. I don't even know what will happen before.

darionyaphet (approved):

PR #392, opened 2021-03-16
Upgrade tool forced type conversion

panda-sheep:
As title


panda-sheep:
@randomJoe211 The document here needs to be modified

src/codec/RowWriterV2.cpp
```diff
@@ -230,6 +230,60 @@ WriteResult RowWriterV2::setValue(folly::StringPiece name, const Value& val) noe
...
+    CHECK(!finished_) << "You have called finish()";
+    auto name = schema_->getFieldName(index);
+    auto propType = schema_->getFieldType(name);
+    return setTypeDefaultValueImp(propType, index);
+}
+
+
+WriteResult
+RowWriterV2::setTypeDefaultValueImp(meta::cpp2::PropertyType& propType, ssize_t index) noexcept {
+    Date d;
```
liuyu85cn, line 251: static this ?

panda-sheep: Yeah，good catch!

Shylock-Hg:
How about the  order of field default value, null value, and *report error*?

bright-starry-sky:
A question, will the upgrade cause the result set to be incorrect? for example: 

- create tag t1 (c1 int)
- insert vertex t1(c1) values 2:(2)
- alter tag t1 change (c1 bool)
- fetch prop on t1 2 ---> the result should be 2
- to upgrade
- fetch prop on t1 2 ---> What is the result set at this point? 2 or default bool value?

CPWstatic:
> A question, will the upgrade cause the result set to be incorrect? for example:
> 
> * create tag t1 (c1 int)
> * insert vertex t1(c1) values 2:(2)
> * alter tag t1 change (c1 bool)
> * fetch prop on t1 2 ---> the result should be 2
> * to upgrade
> * fetch prop on t1 2 ---> What is the result set at this point? 2 or default bool value?

It should be 2 based on our design of multi-version schema.

panda-sheep:
> How about the order of field default value, null value, and _report error_?

Isn't this the logic before?

panda-sheep:
> > A question, will the upgrade cause the result set to be incorrect? for example:
> > 
> > * create tag t1 (c1 int)
> > * insert vertex t1(c1) values 2:(2)
> > * alter tag t1 change (c1 bool)
> > * fetch prop on t1 2 ---> the result should be 2
> > * to upgrade
> > * fetch prop on t1 2 ---> What is the result set at this point? 2 or default bool value?
> 
> It should be 2 based on our design of multi-version schema.

Currently read according to the latest schema

Shylock-Hg:
> > How about the order of field default value, null value, and _report error_?
> 
> Isn't this the logic before?

Yes so why change it.

panda-sheep:
> A question, will the upgrade cause the result set to be incorrect? for example:
> 
> * create tag t1 (c1 int)
> * insert vertex t1(c1) values 2:(2)
> * alter tag t1 change (c1 bool)
> * fetch prop on t1 2 ---> the result should be 2
> * to upgrade
> * fetch prop on t1 2 ---> What is the result set at this point? 2 or default bool value?

Because it is written in accordance with the latest schema, here is to convert 2 into a bool value

panda-sheep:
> > > How about the order of field default value, null value, and _report error_?
> > 
> > 
> > Isn't this the logic before?
> 
> Yes so why change it.
This is more reasonable, mysql is also like this

CPWstatic:
> > A question, will the upgrade cause the result set to be incorrect? for example:
> > 
> > * create tag t1 (c1 int)
> > * insert vertex t1(c1) values 2:(2)
> > * alter tag t1 change (c1 bool)
> > * fetch prop on t1 2 ---> the result should be 2
> > * to upgrade
> > * fetch prop on t1 2 ---> What is the result set at this point? 2 or default bool value?
> 
> Because it is written in accordance with the latest schema, here is to convert 2 into a bool value

I think we should use the accordingly version schema, not the latest one.

critical27:
@CPWstatic @bright-starry-sky I don't know read by which schema is better, the one in data or latest one. See https://jira.nebula-graph.io/browse/NG-336.

src/codec/RowWriterV2.cpp
```diff
@@ -852,9 +903,12 @@ WriteResult RowWriterV2::checkUnsetFields() noexcept {
                                    << ", default value: " << defVal
                                    << ", default value expr: " << field->defaultValue()->toString();
                 }
-            } else {
+            } else if (field->nullable()) {
                 // Set NULL
                 setNullBit(field->nullFlagPos());
+            } else {
+                // Set type default value
+                r = setTypeDefaultValue(i);
```
critical27, line 911: Mark.

critical27:
After discussed with sherman-the-tank, see https://github.com/vesoft-inc/nebula-storage/issues/394.

src/tools/db-upgrade/DbUpgrader.cpp
```diff
@@ -691,11 +690,170 @@ void UpgraderSpace::encodeVertexValue(PartitionID partId,
...
+    }
+}
+
+
 // Used for vertex and edge
 std::string UpgraderSpace::encodeRowVal(const RowReader* reader,
                                         const meta::NebulaSchemaProvider* schema,
                                         std::vector<std::string>& fieldName) {
-    auto oldSchema = reader->getSchema();
+    auto oldSchema = dynamic_cast<const meta::NebulaSchemaProvider*>(reader->getSchema());
```
critical27, line 856: No need to `dynamic_cast` here.

panda-sheep:
two case：

case 1：
create tag player(name string, age int);
insert vertex player(name, age) value 11:("aaaa",1);
insert vertex player(name, age) value 33:("1111", 3);
alter tag player change (name int);
alter tag player add (c string);

result：after the upgrade, 0 records

case 2：
create tag player(name string, age int);
insert vertex player(name, age) value 11:("aaaa",1);
insert vertex player(name, age) value 33:("1111", 3);
alter tag player change (name int);
alter tag player add (c string default “zzz”);

result：after the upgrade, 1 records

critical27 (commented):
Good job

src/tools/db-upgrade/DbUpgrader.cpp
panda-sheep: Yes, I changed the convertValue parameter to SchemaProviderIf*

critical27 (approved):
Good job

liuyu85cn (approved):

PR #403, opened 2021-03-22
Generate log directory, if it does not exist

panda-sheep:
As title

When we start the service with the following script, the log directory will be created in the script.
` ./scripts/nebula.service -c  xxx/etc/nebula-storaged.conf start storaged> /dev/null 2>&1 &`

But when we start the service in other ways, because the directory log does not exist, the startup fails.
`xxx/bin/nebula-storaged --flagfile xxx/etc/nebula-storaged-1.conf >1.log 2 >&1 &`

Therefore, before starting the service, the log directory needs to be created automatically when it does not exist.

### DIFF
--- a/conf/nebula-metad.conf.default
+++ b/conf/nebula-metad.conf.default
@@ -5,7 +5,7 @@
 --pid_file=pids/nebula-metad.pid
 
 ########## logging ##########
-# The directory to host logging files, which must already exists
+# The directory to host logging files
 --log_dir=logs
 # Log level, 0, 1, 2, 3 for INFO, WARNING, ERROR, FATAL respectively
 --minloglevel=0
--- a/conf/nebula-metad.conf.production
+++ b/conf/nebula-metad.conf.production
@@ -5,7 +5,7 @@
 --pid_file=pids/nebula-metad.pid
 
 ########## logging ##########
-# The directory to host logging files, which must already exists
+# The directory to host logging files
 --log_dir=logs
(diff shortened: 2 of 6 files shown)


whitewum:
what...

critical27:
I am not pretty sure if we should do it here. Usually we don't start cluster by `--flagfile` except some cases we don't use the scripts. Maybe do it in chaos?

panda-sheep:
> I am not pretty sure if we should do it here. Usually we don't start cluster by `--flagfile` except some cases we don't use the scripts. Maybe do it in chaos?

Yeah, chaos uses this method.  
The service fails to start because the log directory cannot be created.

panda-sheep:
In fact, I have always been confused. 
why create the log directory in the script when starting the service with the script？
Wouldn't it be better to create the log directory by itself when starting the service program?

critical27:
> In fact, I have always been confused.
> why create the log directory in the script when starting the service with the script？
> Wouldn't it be better to create the log directory by itself when starting the service program?

Yeah, if we create log_dir in program instead of scripts, I think it is ok, just add to graph as well.

panda-sheep:
> log_dir
Agree

panda-sheep:
> > In fact, I have always been confused.
> > why create the log directory in the script when starting the service with the script？
> > Wouldn't it be better to create the log directory by itself when starting the service program?
> 
> Yeah, if we create log_dir in program instead of scripts, I think it is ok, just add to graph as well.
Agree

critical27 (approved):

liuyu85cn (approved):
good jooooob

PR #408, opened 2021-03-24
fix index data loss when insert data during rebuilding index

panda-sheep:
As title.

The steps are as follows:
1) If the index status of the specified space specified part is the building state at this time
In `AddVerticesProcessor::doProcessWithIndex`, the old index data to be deleted and the new index data added are organized into a vector, ready to be added to the operation table.

2) Because `AddVerticesProcessor::doProcessWithIndex` writes vector data into the operation table asynchronously. Release `env_->onFlyingRequest` in `IndexCountWrapper wrapper` before writing.

3) In `RebuildIndexTask::buildIndexOnOperations`, if it is judged that `env_->onFlyingRequest `is 0, it is assumed that the operation table data has been processed, and the rebuild part is completed.

4) At this time, step 2 writes the data into the operation table. At this time, the data written into the part in the operation table will always be invalid, because the data will never be processed.

So this part of the index data is lost.

Thanks chaos for discovering this problem.

### DIFF
--- a/src/storage/mutate/AddEdgesProcessor.cpp
+++ b/src/storage/mutate/AddEdgesProcessor.cpp
@@ -268,9 +268,9 @@ void AddEdgesProcessor::doProcessWithIndex(const cpp2::AddEdgesRequest& req) {
             continue;
         }
         env_->kvstore_->asyncAppendBatch(spaceId_, partId, std::move(batch),
-            [l = std::move(lg), &wrapper, partId, this](kvstore::ResultCode kvRet) {
+            [l = std::move(lg), icw = std::move(wrapper), partId, this](kvstore::ResultCode kvRet) {
                 UNUSED(l);
-                UNUSED(wrapper);
+                UNUSED(icw);
                 handleAsync(spaceId_, partId, kvRet);
             });
     }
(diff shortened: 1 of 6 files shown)


src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -267,8 +268,9 @@ void AddEdgesProcessor::doProcessWithIndex(const cpp2::AddEdgesRequest& req) {
             continue;
         }
         env_->kvstore_->asyncAppendBatch(spaceId_, partId, std::move(batch),
-            [l = std::move(lg), partId, this](kvstore::ResultCode kvRet) {
+            [l = std::move(lg), &wrapper, partId, this](kvstore::ResultCode kvRet) {
```
critical27, line 271: You can't use `&wrapper` here, it is basically like previous before. Because `wrapper` will be released once it is out of scope. Perhaps you need to add a move constructor of `IndexCountWrapper`, just like the `lg` here.

critical27 (commented):
Good job.

src/storage/mutate/AddEdgesProcessor.cpp
panda-sheep: Yeah， 👍

bright-starry-sky (approved):
Well done.

critical27 (approved):
LGTM now, well done

PR #413, opened 2021-03-30
insert vertex & edge if not exists

nevermore3:
depend on https://github.com/vesoft-inc/nebula-common/pull/473


src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -160,25 +179,23 @@ void AddVerticesProcessor::doProcessWithIndex(const cpp2::AddVerticesRequest& re
                 }
 
                 auto key = NebulaKeyUtils::vertexKey(spaceVidLen_, partId, vid, tagId);
+                if (ifNotExists_ && !visited.emplace(key).second) {
```
critical27, line 182: What is the expected order when graph insert same (vId + tagId) with different value?

src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -92,6 +94,20 @@ void AddVerticesProcessor::doProcess(const cpp2::AddVerticesRequest& req) {
                 }
 
                 auto key = NebulaKeyUtils::vertexKey(spaceVidLen_, partId, vid, tagId);
+                if (ifNotExists_) {
+                    if (!visited.emplace(key).second) {
+                        continue;
+                    }
+                    auto obsIdx = findOldValue(partId, vid, tagId);
```
critical27, line 101: Maybe we should emplace result into vertexCache_ in `findOldValue`, at least add a todo

critical27 (commented):
We don't have a good read-modify-write api (asyncAtomicOp could do it, but hard to use) in storage, what is the decision about this `if_not_exists`?

nevermore3:
> We don't have a good read-modify-write api (asyncAtomicOp could do it, but not so friendly) in storage, what is the decision about this `if_not_exists`?

some community users need this feature ,please refer to https://jira.nebula-graph.io/browse/NG-420

src/storage/mutate/AddVerticesProcessor.cpp
nevermore3: in order of input, if encounter the same(vid + tagid) the do nothing when if_not_exists

src/storage/mutate/AddVerticesProcessor.cpp
nevermore3: make sense

bright-starry-sky:
Good job, Three questions： 
1，insert vertex t1(c1) values 1:(1), 1(2), 1(3)... , what is expected to be inserted? 
2, Do we need to consider the index?
3, How about performance?

nevermore3:
> Good job, Three questions：
> 1，insert vertex t1(c1) values 1:(1), 1(2), 1(3)... , what is expected to be inserted?
> 2, Do we need to consider the index?
> 3, How about performance?

1 .  only 1:(1) should be inserted
2. Yes, But the processing of the index is after this logic, so it does not affect the index
3.  In this scenario, performance will inevitably decrease, especially when there is no index.

src/storage/mutate/AddVerticesProcessor.cpp
critical27: So when `ifNotExists_` is true, first one will be inserted, when `ifNotExists_` is false, the last one will be insert. Am I right?

src/storage/mutate/AddVerticesProcessor.cpp
nevermore3: Do you mean vetexCache_->evict(vid,tagId) if (vid,tagId) already exists in kvstroe In the findOldValue function?, i'm a little confused

src/storage/mutate/AddVerticesProcessor.cpp
nevermore3: right

Aiee (dismissed):

src/storage/mutate/AddVerticesProcessor.cpp
critical27: Sorry for late reply, I mean we could insert to cache if we find an old value. The logic could be handled in another PR,  the logic is weird indeed.

nevermore3: It's okay, I understand what you mean

critical27 (approved):
LGTM

CPWstatic (approved):

critical27 (approved):

yixinglu (approved):

critical27 (approved):

liuyu85cn (approved):

PR #417, opened 2021-04-01
Fix bug: when a zone has multiple hosts, zoneName will be moved multi…

guojun85:
…ple times


Shylock-Hg (approved):

laura-ding (approved):
Thanks for your contribution!

darionyaphet (approved):

yixinglu (approved):

PR #420, opened 2021-04-01
Fix core issue: for a new host with no partition, can't find it from …

guojun85:
…hostParts.

Find a core issue, steps to reproduce this issue.
1. create a space on group_0, which contains(zone_0, zone_1, zone_2)
1. Add a new host（127.0.0.1:6778） to zone_0
2. run balance data

as （127.0.0.1:6778）host no part, and can't find it in hostParts, nebula-metad will core at following line:
 zoneParts_[it->first] = ZoneNameAndParts(name, partIter->second);


darionyaphet (approved):
meticulous & thanks

critical27 (approved):

PR #423, opened 2021-04-06
Meta handles leader changes and other errorcodes uniformly

panda-sheep:
For meta, when the leader changes, it is not handled correctly in most cases, and an incorrect error code is returned. So the leader of the meta was changed, but the meta client received other error codes and did not try again.

This pr uniformly handles the error codes of various operations in meta `BaseProcessor and so on`.
benefit:
1) The leader change situation of meta can be processed uniformly .
2) It also facilitates the unified processing of subsequent errorcodes.


critical27:
Good, just remove all `StatusOr` in meta...

src/meta/MetaServiceUtils.cpp
```diff
@@ -905,26 +905,6 @@ std::string MetaServiceUtils::genTimestampStr() {
     return ch;
 }
 
-folly::Optional<bool> MetaServiceUtils::isIndexRebuilding(kvstore::KVStore* kvstore) {
```
critical27, line 908: Maybe don't move them to `BaseProcessor`? some class doesn't inherit from `BaseProcessor`

panda-sheep: yeah， good point. will fix.

bright-starry-sky:
nice job，massive project.

src/meta/ActiveHostsMan.cpp
```diff
@@ -35,20 +36,22 @@ kvstore::ResultCode ActiveHostsMan::updateHostInfo(kvstore::KVStore* kv,
         baton.post();
     });
     baton.wait();
-    return ret;
+    return MetaCommon::to(ret);
```
bright-starry-sky, line 39: I thought, should we completely unified `kvstore::ResultCode` and `cpp2::ErrorCode` ?
For historical reasons, kvstore module cannot depend on storage module, but now that doesn't seem to be the case. so can we using the cpp2::ErrorCode in kvstore module direct?

src/meta/ActiveHostsMan.cpp
```diff
@@ -86,76 +92,98 @@ std::vector<HostAddr> ActiveHostsMan::getActiveHostsInZone(kvstore::KVStore* kv,
...
+ActiveHostsMan::getHostInfo(kvstore::KVStore* kv, const HostAddr& host) {
     auto hostKey = MetaServiceUtils::hostKey(host.host, host.port);
     std::string hostValue;
     auto ret = kv->get(kDefaultSpaceId, kDefaultPartId, hostKey, &hostValue);
     if (ret != kvstore::ResultCode::SUCCEEDED) {
-        LOG(ERROR) << "Get host info " << host << " failed";
-        return Status::Error("Get host info failed");
+        auto retCode = MetaCommon::to(ret);
+        LOG(ERROR) << "Get host info " << host << " failed, error: "
+                   << static_cast<int32_t>(retCode);
```
darionyaphet, line 180: Could turn the code into a string ？

src/meta/ActiveHostsMan.cpp
panda-sheep: Strongly Agree,  👍
At the storage, these two errorcodes are too annoying.

src/meta/ActiveHostsMan.cpp
panda-sheep: good, I will fix it

src/meta/ActiveHostsMan.cpp
panda-sheep: Yeah, this pr is handled like this first, and then I will use a new pr to handle it, otherwise this pr is too big. wdyt?

src/meta/common/MetaCommon.h
```diff
@@ -28,11 +28,14 @@ class MetaCommon final {
         return false;
     }
 
+    // todo unify cpp2::ErrorCode and ResultCode
     static cpp2::ErrorCode to(kvstore::ResultCode code) {
         switch (code) {
         case kvstore::ResultCode::SUCCEEDED:
             return cpp2::ErrorCode::SUCCEEDED;
+        case kvstore::ResultCode::ERR_SPACE_NOT_FOUND:
```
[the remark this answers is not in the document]
panda-sheep: For the client, it sees cpp2::ErrorCode
So there is no need to distinguish here

critical27:
Conflict again, maybe because there are several reverts these days. You could contact @liuyu85cn if necessary.

panda-sheep:
test passed, ready to review.

critical27 (approved):
Good job. LGTM

liuyu85cn (approved):
LGTM

PR #425, opened 2021-04-07
Fix docker tag according to branch name

yixinglu:
as title.


yixinglu:
It's meaningless to build docker image each time of every commit to release branch.

PR #428, opened 2021-04-09
restore update atomic operation

panda-sheep:
update is read-modify-write, which is an atomic operation.
 pr https://github.com/vesoft-inc/nebula-storage/pull/355 change update operation into a non-atomic operation.
The result of the current update is unstable.

According to the logic before this modification, the following case:
```
create tag course(credits int);
insert vertex course(credits) VALUES "200":(1);
```
Then the two clients are executed separately
```
UPSERT VERTEX "200" SET course.credits = 2 WHEN $^.course.credits == 1 YIELD $^.course.credits;
UPSERT VERTEX "200" SET course.credits = 3 WHEN $^.course.credits == 1 YIELD $^.course.credits;
```
The second update statement may be executed successfully or failed.

The expectations are:
The execution of the second update statement failed.

### DIFF
--- a/src/storage/exec/UpdateNode.h
+++ b/src/storage/exec/UpdateNode.h
@@ -167,10 +167,24 @@ public:
 
     kvstore::ResultCode execute(PartitionID partId, const VertexID& vId) override {
         CHECK_NOTNULL(planContext_->env_->kvstore_);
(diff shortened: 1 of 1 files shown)


critical27 (approved):
LGTM

liuyu85cn (approved):
LGTM

PR #430, opened 2021-04-12
Fix cleanWAL block raft heartbeat

critical27:
Sad story. After long time writing data, there will be plenty of wals, and `NebulaStore::cleanWAL` will take way more time (> 10s), which will block raft heartbeat because they share the same worker. So in this PR:
1. separate background worker thread.
2. clean some unused wal parameters in 2.0
3. fix endian in rebuild index


panda-sheep (approved):
Good job

darionyaphet (approved):
LGTM

liuyu85cn (approved):
Good job

PR #433, opened 2021-04-14
Convert pair to tuple about ttlProps

darionyaphet:


bright-starry-sky:
What is the purpose of this PR?  In general, std::pair faster than std::tuple.

darionyaphet:
> What is the purpose of this PR? In general, std::pair faster than std::tuple.

the nested structure std::pair<key, std::pair<>> is really faster than a flat structure？I don't know. But it's strange so I fix it.

bright-starry-sky:
> > What is the purpose of this PR? In general, std::pair faster than std::tuple.
> 
> the nested structure std::pair<key, std::pair<>> is really faster than a flat structure？I don't know. But it's strange so I fix it.

Emm, when the pair.first is false, the pair.second can be empty.

darionyaphet:
> > > What is the purpose of this PR? In general, std::pair faster than std::tuple.
> > 
> > 
> > the nested structure std::pair<key, std::pair<>> is really faster than a flat structure？I don't know. But it's strange so I fix it.
> 
> Emm, when the pair.first is false, the pair.second can be empty.

non existent although `!(duration <= 0 || col.empty())` is false,

src/storage/CommonUtils.cpp
```diff
@@ -114,8 +114,8 @@ bool CommonUtils::checkDataExpiredForTTL(const meta::SchemaProviderIf* schema,
     return false;
 }
 
-std::pair<bool, std::pair<int64_t, std::string>>
-    CommonUtils::ttlProps(const meta::SchemaProviderIf* schema) {
+std::tuple<bool, int64_t, std::string>
```
monadbobo, line 117: It is more intuitive to use std::optional<std::tuple<int64_t, std::string>> here.

PR #438, opened 2021-04-14
Raft heartbeat process in event base

critical27:
Our workflow looks like below, as for an write request

| leader bgWorker | leader ioThreadPool | leader threadMananger | follower ioThreadPool | follower threadManager |
| --------------- | ------------------- | --------------------- | --------------------- | ---------------------- |
|                 | 1. requestReceived     |                       |                       |                        |
|                 |                     | 2. processInThread       |                       |                        |
| (heartbeat)     | 3. replicate           |                       |                       |                        |
|                 |                     |                       | 4. receive appendLog       |                        |
|                 |                     |                       |                       | 5. process      |
|                 | 6. collectN            |                       |                       |                        |
|                 |                     | 7. commit                |                       |                        |

From the view of a storaged process, there would be many partition, we could be blocked in **any** phase if it takes longer time to process. Phase 5 and 7 are the main cause of blocking as before. When pressure is big enough, it is possible that all worker thread is busy. This would raise a lot problems, one of the most notorious among them is leader change. 

Block reason:
1. lock
2. rocksdb write stall (while holding lock)

IMO, there should be only one phase could be blocked, which is leader commit (phase 7). To achieve that, many works need to do:
1. The second commit `don not hold raft lock when commit` will relax the restriction of raft lock. (we have another lock replicatingLogs to prevent concurrent replicate)
2. The third commit `follower delay commit if write stall` will make follower commit logs not to block (phase 5, follower could commit in async)
3. The first commit `heartbeat refactor` make heartbeat process in event base, even if all worker threads are blocked, there won't be unexpected election.

So, the heartbeat logic is:
1. If a log is on the fly, only send a heartbeat which will only be processed in eventbase.
2. If no log is on the fly, both a dummy log(previous behavior) and a heartbeat will be sent.

depends on https://github.com/vesoft-inc/nebula-common/pull/497

### DIFF
--- a/src/kvstore/raftex/Host.cpp
+++ b/src/kvstore/raftex/Host.cpp
@@ -508,6 +508,7 @@ folly::Future<cpp2::HeartbeatResponse> Host::sendHeartbeat(folly::EventBase* eb,
     req->set_last_log_term_sent(lastLogTerm);
     req->set_last_log_id_sent(lastLogId);
     folly::Promise<cpp2::HeartbeatResponse> promise;
+    auto future = promise.getFuture();
     sendHeartbeatRequest(eb, std::move(req))
         .via(eb)
         .then([self = shared_from_this(), pro = std::move(promise)]
@@ -522,7 +523,7 @@ folly::Future<cpp2::HeartbeatResponse> Host::sendHeartbeat(folly::EventBase* eb,
                 pro.setValue(std::move(t.value()));
             }
         });
-    return promise.getFuture();
+    return future;
 }
 
 folly::Future<cpp2::HeartbeatResponse> Host::sendHeartbeatRequest(
--- a/src/kvstore/raftex/Host.cpp
+++ b/src/kvstore/raftex/Host.cpp
@@ -151,11 +151,6 @@ folly::Future<cpp2::AppendLogResponse> Host::appendLogs(
                       << ", lastLogIdSent = " << lastLogIdSent_
                       << ", lastLogTermSent = " << lastLogTermSent_;
         }
-        if (prevLogTerm < lastLogTermSent_ || prevLogId < lastLogIdSent_) {
-            LOG(INFO) << idStr_ << "We have sended this log, so go on from id " << lastLogIdSent_
-                      << ", term " << lastLogTermSent_ << "; current prev log id " << prevLogId
-                      << ", current prev log term " << prevLogTerm;
-        }
         logTermToSend_ = term;
         logIdToSend_ = logId;
         committedLogId_ = committedLogId;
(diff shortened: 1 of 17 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1680,36 +1648,46 @@ void RaftPart::processAppendLogRequest(
...
-    if (req.get_committed_log_id() > committedLogId_) {
+    LogID lastLogIdCanCommit = std::min(lastLogId_, req.get_committed_log_id());
+    if (lastLogIdCanCommit > committedLogId_) {
         // Commit some logs
         // We can only commit logs from firstId to min(lastLogId_, leader's commit log id),
         // follower can't always commit to leader's commit id because of lack of log
-        LogID lastLogIdCanCommit = std::min(lastLogId_, req.get_committed_log_id());
-        CHECK_LE(committedLogId_ + 1, lastLogIdCanCommit);
-        if (commitLogs(wal_->iterator(committedLogId_ + 1, lastLogIdCanCommit))) {
+        auto code = commitLogs(wal_->iterator(committedLogId_ + 1, lastLogIdCanCommit), false);
```
bright-starry-sky, line 1656: One question, when` wait == false` , we will ignore whether the log commits successfully or not.but why do you need to set the committedLogId_ at line 1671?

src/kvstore/raftex/Host.cpp
```diff
@@ -494,18 +478,83 @@ folly::Future<cpp2::AppendLogResponse> Host::sendAppendLogRequest(
...
+            if (t.hasException()) {
+                cpp2::HeartbeatResponse resp;
+                resp.set_error_code(cpp2::ErrorCode::E_EXCEPTION);
+                pro.setValue(std::move(resp));
+                return;
+            } else {
+                pro.setValue(std::move(t.value()));
+            }
+        });
+    return promise.getFuture();
```
liuyu85cn, line 525: may be moved(not sure about this)?

src/kvstore/raftex/RaftPart.cpp
critical27: L1671 only set committedLogId_ in resp, but we didn't update the `committedLogId_`, when we commit successfully, we update the committedLogId_, and response the new one (Line 1661 1662).

bright-starry-sky (approved):

liuyu85cn (approved):

src/kvstore/raftex/Host.cpp
critical27: fixed

PR #442, opened 2021-04-15
base toss

liuyu85cn:
This PR is to address this issue that in-edge and out-edge may in-consist while add and update. 

I add an upper tier above the normal AddEdgeProcess / UpdateEdgeProcessor. 

and I named them Chain...Processor. 

spaces who set TOSS on will call those chain processor, and the chain processor will drive the normal ones 

to do the actually insert/update.


darionyaphet:
WHAT IS THIS ?

src/storage/transaction/ChainAddEdgesProcessorLocal.cpp
```diff
@@ -0,0 +1,357 @@
...
+        }
+        onFinished();
+        return;
+    }
+
+    spaceVidLen_ = ret.value();
+
+    if (type != ChainProcessType::NORMAL) {
+        std::vector<std::string> keys = strEdgekeys(req);
+        edgeVers_ = ConsistUtil::getMultiEdgeVers(env_->kvstore_, spaceId, localPartId_, keys);
```
critical27, line 64: Where do we use `edgeVers_`? Because I see `ChainAddEdgesProcessorRemote::checkVersion` is checking this version to prevent overwrite newer data.

src/storage/transaction/ChainResumeProcessor.cpp
```diff
@@ -0,0 +1,111 @@
...
+        LOG(INFO) << "no leader found, skip any resume process";
+        return;
+    }
+    std::unique_ptr<kvstore::KVIterator> iter;
+    for (auto& leader : leaders) {
+        auto spaceId = leader.first;
+        LOG(INFO) << "leader.second.size() = " << leader.second.size();
+        for (auto& partInfo : leader.second) {
+            auto partId = partInfo.get_part_id();
+            auto prefix = ConsistUtil::primePrefix(partId);
```
critical27, line 28: How about add a timestamp after partId for primePrefix? Because in worst case for every insert/update, the resueme will be triggered. (doublePrimePrefix is ok because it is not written in normal insert/update path)

src/storage/transaction/ChainUpdateEdgeProcessorLocal.cpp
```diff
@@ -0,0 +1,295 @@
...
+    return processNormalLocal(code);
+}
+
+folly::SemiFuture<ECode>
+ChainUpdateEdgeProcessorLocal::processResumeRemoteLocal(ECode code) {
+    LOG(INFO) << __func__ << "(), code = " << apache::thrift::util::enumNameSafe(code);
+    if (code == ECode::SUCCEEDED) {
+        auto partId = req_.get_part_id();
+        auto key = ConsistUtil::doublePrime(spaceVidLen_, partId, req_.get_edge_key());
+        kvErased_.emplace_back(key);
```
critical27, line 164: Am I missing something here? I think double prime has been add to `kvErased` in `prepareLocal` around line 80

src/storage/exec/UpdateNode.h
```diff
@@ -765,6 +751,14 @@ class UpdateEdgeNode : public UpdateNode<cpp2::EdgeKey> {
         }
         // step 3, insert new edge data
         batchHolder->put(std::move(key_), std::move(nVal));
+
+        // extra phase: if there are some extra requirement.
+        for (auto& [k, v] : edgeContext_->kvAppend) {
```
critical27, line 756: Make sure `kvAppend` and `kvErased` doen't have overlaps here. Because `doublePrime` in `ChainUpdateEdgeProcessorLocal` could be add to both of them here.

src/storage/transaction/GetValueProcessor.cpp
```diff
@@ -21,7 +21,7 @@ void GetValueProcessor::process(const cpp2::GetValueRequest& req) {
 
     std::string value;
     auto rc = env_->kvstore_->get(spaceId, partId, key, &value);
-    LOG_IF(INFO, FLAGS_trace_toss) << "getValue for partId=" << partId
+    LOG_IF(INFO, FLAGS_trace_toss) << "getValue() partId=" << partId
```
critical27, line 24: We don't need this class for now?

critical27 (commented):
Huge PR, good job! Some details need to be discussed, see the comments. 

We could unify some code in ChainUpdate and ChainAdd later.

src/meta/processors/partsMan/GetPartsAllocProcessor.cpp
```diff
@@ -34,9 +34,65 @@ void GetPartsAllocProcessor::process(const cpp2::GetPartsAllocReq& req) {
...
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto rc = kvstore_->prefix(kDefaultSpaceId, kDefaultPartId, prefix, &iter);
+    if (rc != nebula::cpp2::ErrorCode::SUCCEEDED) {
+        LOG(ERROR) << "Access kvstore failed, spaceId " << spaceId
+                   << apache::thrift::util::enumNameSafe(rc);
+        return ret;
+    }
+
+    std::vector<PartitionID> partIdVec;
+    std::vector<std::string> keys;
```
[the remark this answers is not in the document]
liuyu85cn: Mmmmm... It's used to scan all part of a space. Maybe there won't be any dup?

src/storage/transaction/ChainUpdateEdgeProcessorLocal.cpp
liuyu85cn: good catch

src/storage/exec/UpdateNode.h
liuyu85cn: either?

critical27: Basically, you won't either write `doublePrime` or remove `doublePrime`, but if the batch contains both `write` and `remove`, how can we make sure which happens first?

src/meta/processors/partsMan/GetPartsAllocProcessor.cpp
```diff
@@ -34,9 +34,65 @@ void GetPartsAllocProcessor::process(const cpp2::GetPartsAllocReq& req) {
...
     handleErrorCode(nebula::cpp2::ErrorCode::SUCCEEDED);
     resp_.set_parts(std::move(parts));
+    auto terms = getTerm(spaceId);
+    if (!terms.empty()) {
+        resp_.set_terms(std::move(terms));
+    }
     onFinished();
 }
 
+std::unordered_map<PartitionID, TermID> GetPartsAllocProcessor::getTerm(GraphSpaceID spaceId) {
```
bright-starry-sky, line 44: A suggestion，is better to change the argument to `const vector<std::string>& keys`? this will reduce a database scan of `MetaServiceUtils::partPrefix(spaceId)`.

src/meta/processors/partsMan/GetPartsAllocProcessor.cpp
```diff
@@ -34,9 +34,65 @@ void GetPartsAllocProcessor::process(const cpp2::GetPartsAllocReq& req) {
...
+    LOG(INFO) << "keys.size() = " << keys.size();
+    auto [code, statusVec] = kvstore_->multiGet(0, 0, std::move(keys), &vals);
+    if (code != nebula::cpp2::ErrorCode::SUCCEEDED &&
+        code != nebula::cpp2::ErrorCode::E_PARTIAL_RESULT) {
+        LOG(INFO) << "error rc = " << apache::thrift::util::enumNameSafe(code);
+        return ret;
+    }
+
+    TermID term;
+    for (auto i = 0U; i != keys.size(); ++i) {
```
bright-starry-sky, line 80: is better instead of keys to vals?

src/storage/exec/EdgeNode.h
```diff
@@ -137,15 +135,8 @@ class FetchEdgeNode final : public EdgeNode<cpp2::EdgeKey> {
...
-                context_->env()->txnMan_->enableToss(context_->spaceId())) {
-                bool stopAtFirstEdge = true;
-                iter_.reset(new TossEdgeIterator(
-                    context_, std::move(iter), edgeType_, schemas_, &ttl_, stopAtFirstEdge));
-            } else {
-                iter_.reset(new SingleEdgeIterator(
-                    context_, std::move(iter), edgeType_, schemas_, &ttl_, false));
-            }
+            iter_.reset(new SingleEdgeIterator(
+                        context_, std::move(iter), edgeType_, schemas_, &ttl_, false));
```
bright-starry-sky, line 139: is this gflag toss will be deleted?

liuyu85cn: yes

src/meta/processors/partsMan/GetPartsAllocProcessor.cpp
liuyu85cn: you mean keys may move away?

bright-starry-sky: > you mean keys may move away?

Yes, that's what I think.

liuyu85cn: good catch, will fix it.

src/meta/processors/partsMan/GetPartsAllocProcessor.cpp
```diff
@@ -34,9 +34,113 @@ void GetPartsAllocProcessor::process(const cpp2::GetPartsAllocReq& req) {
...
+    if (!nebula::ok(spaceVal)) {
+        auto rc = nebula::error(spaceVal);
+        LOG(ERROR) << "Get Space SpaceId: " << spaceId
+                   << " error: " << apache::thrift::util::enumNameSafe(rc);
+        handleErrorCode(rc);
+        return ret;
+    }
+
+    auto properties = MetaServiceUtils::parseSpace(nebula::value(spaceVal));
+    auto partNum = properties.get_partition_num();
```
bright-starry-sky, line 58: 👍

PR #446, opened 2021-04-20
fix 1 replica balance failed, rebuild index ut

critical27:
Fixes https://github.com/vesoft-inc/nebula-graph/issues/969

### DIFF
--- a/src/storage/admin/AdminProcessor.h
+++ b/src/storage/admin/AdminProcessor.h
@@ -58,7 +58,6 @@ public:
         if (partMeta.hosts_.size() == 1) {
             LOG(INFO) << "Skip transfer leader of space " << spaceId
                       << ", part " << partId << " because of single replica.";
-            this->pushResultCode(cpp2::ErrorCode::E_TRANSFER_LEADER_FAILED, partId);
             onFinished();
             return;
         }
--- a/src/storage/test/RebuildIndexTest.cpp
+++ b/src/storage/test/RebuildIndexTest.cpp
@@ -406,8 +406,7 @@ TEST_F(RebuildIndexTest, RebuildEdgeIndexWithDelete) {
         auto* processor = DeleteEdgesProcessor::instance(RebuildIndexTest::env_, nullptr);
         auto fut = processor->getFuture();
         processor->process(req);
-        auto resp = std::move(fut).get();
-        EXPECT_EQ(0, resp.result.failed_parts.size());
+        std::move(fut).get();
     };
 
     // Add Edges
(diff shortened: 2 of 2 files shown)


src/storage/admin/AdminProcessor.h
```diff
@@ -58,7 +58,6 @@ class TransLeaderProcessor : public BaseProcessor<cpp2::AdminExecResp> {
         if (partMeta.hosts_.size() == 1) {
             LOG(INFO) << "Skip transfer leader of space " << spaceId
                       << ", part " << partId << " because of single replica.";
-            this->pushResultCode(cpp2::ErrorCode::E_TRANSFER_LEADER_FAILED, partId);
```
randomJoe211, line 61: Why remove the error msg? The user should know if the BALANCE operation worked. And if the operation failed, we should provide a clear cause like "To balance the xx, the graph spaces must have more than one replicas."

src/storage/test/RebuildIndexTest.cpp
```diff
@@ -406,8 +406,7 @@ TEST_F(RebuildIndexTest, RebuildEdgeIndexWithDelete) {
         auto* processor = DeleteEdgesProcessor::instance(RebuildIndexTest::env_, nullptr);
         auto fut = processor->getFuture();
         processor->process(req);
-        auto resp = std::move(fut).get();
-        EXPECT_EQ(0, resp.result.failed_parts.size());
```
bright-starry-sky, line 410: A good way.

src/storage/admin/AdminProcessor.h
```diff
@@ -58,7 +58,6 @@ class TransLeaderProcessor : public BaseProcessor<cpp2::AdminExecResp> {
         if (partMeta.hosts_.size() == 1) {
             LOG(INFO) << "Skip transfer leader of space " << spaceId
```
darionyaphet, line 59: It seems not allowed to balance for the single replica cluster. 

Should we support some warning before the balance be executed?

critical27: Why forbid it, it is legal

src/storage/admin/AdminProcessor.h
critical27: This is way more complicated than you thought. Actually, there won't be this phase for 1 replica balance

darionyaphet: the single replica may not require transfer leader so it's fine @randomJoe211

bright-starry-sky (approved):
LGTM

darionyaphet (approved):

PR #447, opened 2021-04-20
Support config about index limit

darionyaphet:


bright-starry-sky:
There is one possible scenario that we need to consider，what is the maximum number of columns supported?
The nullable bitSet(2 byte) contained in the index key, maximum of 16 columns are allowed.

darionyaphet:
> There is one possible scenario that we need to consider，what is the maximum number of columns supported?
> The nullable bitSet(2 byte) contained in the index key, maximum of 16 columns are allowed.

skillfully

If so, The number of index columns must not be greater than the nullable column size...

bright-starry-sky:
> > There is one possible scenario that we need to consider，what is the maximum number of columns supported?
> > The nullable bitSet(2 byte) contained in the index key, maximum of 16 columns are allowed.
> 
> skillfully
> 
> If so, The number of index columns must not be greater than the nullable column size...

Actual, columns that are nullable must be less than or equal to 16, otherwise, else we need to increase that `nullable bitSet`.
Also, for the sake of simplicity, we currently limit the column maximum to 16 without checking whether nullable.

darionyaphet:
> > > There is one possible scenario that we need to consider，what is the maximum number of columns supported?
> > > The nullable bitSet(2 byte) contained in the index key, maximum of 16 columns are allowed.
> > 
> > 
> > skillfully
> > If so, The number of index columns must not be greater than the nullable column size...
> 
> Actual, columns that are nullable must be less than or equal to 16, else we need to increase that `nullable bitSet`.
> Also, for the sake of simplicity, we currently limit the column maximum to 16 without checking whether nullable.

It looks fine & 16 for the nullable column is enough.  I will fix the test case and avoid hard code.

bright-starry-sky:
> > > > There is one possible scenario that we need to consider，what is the maximum number of columns supported?
> > > > The nullable bitSet(2 byte) contained in the index key, maximum of 16 columns are allowed.
> > > 
> > > 
> > > skillfully
> > > If so, The number of index columns must not be greater than the nullable column size...
> > 
> > 
> > Actual, columns that are nullable must be less than or equal to 16, else we need to increase that `nullable bitSet`.
> > Also, for the sake of simplicity, we currently limit the column maximum to 16 without checking whether nullable.
> 
> It looks fine & 16 for the nullable column is enough. I will fix the test case and avoid hard code.

👍

monadbobo (dismissed):
LGTM.

critical27 (approved):
LGTM

bright-starry-sky (approved):
Well done.

PR #454, opened 2021-04-27
fix show hosts errors

panda-sheep:
As title~
1) Because the corresponding leader information was not deleted when the space was dropped, then the leader information was incorrect when `show host`.
2) Even if the corresponding leader information is deleted during `drop space`, the` show hosts` result  will occasionally be incorrect. The reasons are as follows:

When dropping the space，heartBeat (HBProcessor) will also write leader information of this space to the meta.
As a result, after the drop space, the leader information written by heartBeat may be out of date. 
Cause `show hosts`  to display incorrect leader information.

Two verification methods：
1）when no any space，
```
create space yy(partition_num=1, replica_factor=1);
show hosts;
drop space yy;
show hosts;
```
2)  After `drop space`, there is no space, use meta_dump tool

### DIFF
--- a/src/meta/processors/partsMan/DropSpaceProcessor.cpp
+++ b/src/meta/processors/partsMan/DropSpaceProcessor.cpp
@@ -101,25 +101,6 @@ void DropSpaceProcessor::process(const cpp2::DropSpaceReq& req) {
     auto statiskey = MetaServiceUtils::statisKey(spaceId);
     deleteKeys.emplace_back(statiskey);
 
-    // 6. Delete related leader data
-    auto leaderKey = MetaServiceUtils::leaderPrefix(spaceId);
-
-    auto leaderRet = doPrefix(leaderKey);
-    if (!nebula::ok(leaderRet)) {
-        auto retCode = nebula::error(leaderRet);
-        LOG(ERROR) << "Drop space Failed, space " << spaceName
-                   << " error: " << apache::thrift::util::enumNameSafe(retCode);
-        handleErrorCode(retCode);
-        onFinished();
-        return;
-    }
-
-    auto leaderIter = nebula::value(leaderRet).get();
-    while (leaderIter->valid()) {
-        deleteKeys.emplace_back(leaderIter->key());
-        leaderIter->next();
-    }
-
     doSyncMultiRemoveAndUpdate(std::move(deleteKeys));
     LOG(INFO) << "Drop space " << spaceName << ", id " << spaceId;
 }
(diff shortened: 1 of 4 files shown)


src/meta/processors/partsMan/DropSpaceProcessor.cpp
```diff
@@ -96,10 +97,29 @@ void DropSpaceProcessor::process(const cpp2::DropSpaceReq& req) {
         lstIter->next();
     }
 
-    // Delete statis data if it exists
+    // 5. Delete related statis data
     auto statiskey = MetaServiceUtils::statisKey(spaceId);
     deleteKeys.emplace_back(statiskey);
 
+    // 6. Delete related leader data
```
critical27, line 104: IMO, we don't need this step, even if we delete here, subsequent HB will write the leader key as well. Just delete them in `show hosts` is enough.

src/meta/processors/partsMan/ListHostsProcessor.cpp
```diff
@@ -177,10 +177,34 @@ cpp2::ErrorCode ListHostsProcessor::fillLeaders() {
...
+        auto spaceId = spaceIdAndPartId.first;
+        auto partId = spaceIdAndPartId.second;
+
+        // The space id in the leader key may no exist.
+        auto spaceIter = spaceIdNameMap_.find(spaceId);
+        if (spaceIter == spaceIdNameMap_.end()) {
+            removeLeadersKey.emplace_back(iter->key());
+            continue;
+        }
+        // If part invalid , remove leader key
```
critical27, line 195: Maybe just check the space is enough, we don't need to check part again, `drop space` will make sure that all space related will be dropped in an atomic batch. 

Besides, there could be plenty partition, read each part may become a problem.

panda-sheep: Yeah, the `part ` don't need to be checked.
Because according to the format of the leader key, when the spaceId is valid, the `part `must be valid too.

Good catch, 👍

liuyu85cn (approved):
Good Job

bright-starry-sky:
LGTM.
One extra question, why didn't delete the leader info together when drop the space?

panda-sheep:
> didn't
When dropping the space，heartBeat (HBProcessor) will also write leader information of this space to the meta.
As a result, after the drop space, the leader information written by heartBeat may be out of date.

Delete the leader information when you drop space, and write an expired leader information in the later HB. So deleting the leader information when drop space is useless

bright-starry-sky:
> > didn't
> > When dropping the space，heartBeat (HBProcessor) will also write leader information of this space to the meta.
> > As a result, after the drop space, the leader information written by heartBeat may be out of date.
> 
> Delete the leader information when you drop space, and write an expired leader information in the later HB. So deleting the leader information when drop space is useless

Good.

bright-starry-sky (approved):
LGTM

Issue #458, opened 2021-05-11
storage: BALANCE DATA failed with Transfer leader failed, status Unknown code -46


veezhang:
## BALANCE DATA 1620708035
```
 BALANCE DATA 1620708035
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| balanceId, spaceId:partId, src->dst                                                                                                                                                                         | status   |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| "[1620708035, 1:1, test1-1-3-storaged-1.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779->test1-1-3-storaged-3.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779]" | "FAILED" |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| "[1620708035, 1:2, test1-1-3-storaged-2.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779->test1-1-3-storaged-3.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779]" | "FAILED" |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| "[1620708035, 1:3, test1-1-3-storaged-0.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779->test1-1-3-storaged-3.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779]" | "FAILED" |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| "[1620708035, 1:5, test1-1-3-storaged-2.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779->test1-1-3-storaged-3.test1-1-3-storaged-headless.nebulacluster-5088.svc.cluster.local:9779]" | "FAILED" |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
| "Total:4, Succeeded:0, Failed:4, In Progress:0, Invalid:0"                                                                                                                                                  | 0.0      |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------+
```

## meta log
```
[...]

PR #459, opened 2021-05-12
Check host exist or not before remove it

darionyaphet:


critical27:
Maybe take a look at #457 and #458. When shrink cluster, the plan cannot be generated correctly. I believe when `genTasks`, some check is wrong.

darionyaphet:
> Maybe take a look at #457 and #458. When shrink cluster, the plan cannot be generated correctly. I believe when `genTasks`, some check is wrong.

port is break. not a problem

darionyaphet:
> > Maybe take a look at #457 and #458. When shrink cluster, the plan cannot be generated correctly. I believe when `genTasks`, some check is wrong.
> 
> port is break. not a problem

reference :

https://discuss.nebula-graph.com.cn/t/topic/3840/20

critical27:
> > Maybe take a look at #457 and #458. When shrink cluster, the plan cannot be generated correctly. I believe when `genTasks`, some check is wrong.
> 
> port is break. not a problem

Port is not the point. 😂

PR #460, opened 2021-05-13
Support specified columns for full-text index

bright-starry-sky:
1, Changed full-text index structure, deleted field schema_id
2, Support for creating full-text index against fields
3, Support drop full-text index
4, Support show full-text index
5, Refuse to create full-text index if fixed_string type length > 256 
6, Refuse to drop or alter tag if full-text index exists.

depends on vesoft-inc/nebula-common#530

### DIFF
--- a/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
+++ b/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
@@ -133,6 +133,8 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
                 LOG(ERROR) << "Unsupport index type lengths greater than "
                            << MAX_INDEX_TYPE_LENGTH << " : "
                            << field.get_name();
+                handleErrorCode(nebula::cpp2::ErrorCode::E_UNSUPPORTED);
+                onFinished();
                 return;
             }
         } else if (col.type.get_type() == meta::cpp2::PropertyType::STRING) {
--- a/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
+++ b/src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
@@ -4,6 +4,7 @@
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
 
+#include "common/base/CommonMacro.h"
 #include "meta/processors/indexMan/CreateEdgeIndexProcessor.h"
 
 namespace nebula {
@@ -127,10 +128,25 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
             return;
         }
         cpp2::ColumnDef col = *iter;
-        if (col.type.get_type() == meta::cpp2::PropertyType::STRING) {
+        if (col.type.get_type() == meta::cpp2::PropertyType::FIXED_STRING) {
+            if (*col.type.get_type_length() > MAX_INDEX_TYPE_LENGTH) {
+                LOG(ERROR) << "Unsupport index type lengths greater than "
+                           << MAX_INDEX_TYPE_LENGTH << " : "
+                           << field.get_name();
+                return;
+            }
+        } else if (col.type.get_type() == meta::cpp2::PropertyType::STRING) {
             if (!field.type_length_ref().has_value()) {
                 LOG(ERROR) << "No type length set : " << field.get_name();
                 handleErrorCode(nebula::cpp2::ErrorCode::E_INVALID_PARM);
+                onFinished();
+                return;
+            }
+            if (*field.get_type_length() > MAX_INDEX_TYPE_LENGTH) {
+                LOG(ERROR) << "Unsupport index type lengths greater than "
+                           << MAX_INDEX_TYPE_LENGTH << " : "
+                           << field.get_name();
+                handleErrorCode(nebula::cpp2::ErrorCode::E_UNSUPPORTED);
                 onFinished();
                 return;
             }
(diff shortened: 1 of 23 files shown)


bright-starry-sky:
Ready to review.

src/meta/processors/indexMan/FTIndexProcessor.cpp
```diff
@@ -0,0 +1,172 @@
...
+                   << apache::thrift::util::enumNameSafe(retCode);
+        handleErrorCode(retCode);
+        onFinished();
+        return;
+    }
+    auto it = nebula::value(ret).get();
+    // If already have a full-text index that depend on this tag or edge,  reject to create it.
+    // even though it's a different column.
+    while (it->valid()) {
+        auto indexName = MetaServiceUtils::parsefulltextIndexName(it->key());
```
critical27, line 101: If we forbid a schema has more than one FTIndex, the name of FTIndex seems meaningless? Why not just use tag name or edge name?

BTW, consider change the logic from `CreateFTIndex` to `add one column to FTIndex`. IMO, the forbid here is a little aggressive, and it will be hard to extend later. For example, if we want to add another column, we could only drop then recreate. Actually, we only need to rebuild FTIndex of this column.

bright-starry-sky: > If we forbid a schema has more than one FTIndex, the name of FTIndex seems meaningless? Why not just use tag name or edge name?
> 
Good question, I think it's more flexible of this way. The users can name it to `tag name` or `edge name`,  also named another. either way, we can show them through `SHOW FULLTEXT INDEXES`. 
```
(root@nebula) [ft_test_space]> show fulltext indexes
+-----------------+-------------+-------------+----------------------+
| Name            | Schema Type | Schema Name | Fields               |
+-----------------+-------------+-------------+----------------------+
| "nebula_ft_idx" | "Tag"       | "player"    | "c_fixed_string_256" |
+-----------------+-------------+-------------+----------------------+
```

> BTW, consider change the logic from `CreateFTIndex` to `add columns to FTIndex`. IMO, the forbid here is a little aggressive, and it will be hard to extend later. For example, if we want to add another column, we could only drop then recreate. Actually, we only need to rebuild FTIndex of this column.
two ways : 1, drop ftindex -> recreate ftindex -> rebuild ftindex; 2, alter ftindex add column -> rebuild ftindex. Actually I prefer the second one. but the current implementation is the first one, the reason is that we don't support `REBUILD` yet. I think we can do it after `REBUILD` ready. WDYT？

critical27: >  I think we can do it after REBUILD ready. WDYT？
Could consider implement `add columns to FTIndex` here, it won't conflict with this PR. Just forbid upcoming alter. Could do it later as well.

bright-starry-sky: OK , Let's do that in the next PR, the current PR is too big.

critical27 (approved):
LGTM, good job

src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
```diff
@@ -126,13 +127,28 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
             return;
         }
         cpp2::ColumnDef col = *iter;
-        if (col.type.get_type() == meta::cpp2::PropertyType::STRING) {
+        if (col.type.get_type() == meta::cpp2::PropertyType::FIXED_STRING) {
+            if (*col.type.get_type_length() > MAX_INDEX_TYPE_LENGTH) {
+                LOG(ERROR) << "Unsupport index type lengths greater than "
+                           << MAX_INDEX_TYPE_LENGTH << " : "
+                           << field.get_name();
```
panda-sheep, line 134: handle  `handleErrorCode  `and  `onFinished`

src/meta/processors/partsMan/DropSpaceProcessor.cpp
```diff
@@ -101,6 +101,22 @@ void DropSpaceProcessor::process(const cpp2::DropSpaceReq& req) {
...
+        auto retCode = nebula::error(ftRet);
+        LOG(ERROR) << "Drop space Failed, space " << spaceName
+                   << " error: " << apache::thrift::util::enumNameSafe(retCode);
+        handleErrorCode(retCode);
+        onFinished();
+        return;
+    }
+    auto ftIter = nebula::value(ftRet).get();
+    while (ftIter->valid()) {
+        deleteKeys.emplace_back(ftIter->key());
```
panda-sheep, line 116: It should be judged whether the current full-text index record is the current space, right?

panda-sheep (commented):
Good job！

src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
bright-starry-sky: > handle `handleErrorCode `and `onFinished`

Good point. thanks. fixed.

src/meta/processors/partsMan/DropSpaceProcessor.cpp
bright-starry-sky: > It should be judged whether the current full-text index record is the current space, right?

An embarrassing bug. thanks. fixed

critical27 (approved):

panda-sheep (approved):
Excellent！

bright-starry-sky:
ASAN error occur. in order not to affect the overall compilation , let merge this PR and address the ASAN error immediate.

PR #462, opened 2021-05-14
Support readahead_size when range and prefix scan

darionyaphet:
The main purpose is to reduce the scanning time of full data such as rebuild index and statistics

need some benchmark to explain the performance improvement.

### DIFF
--- a/src/kvstore/RocksEngine.cpp
+++ b/src/kvstore/RocksEngine.cpp
@@ -15,6 +15,9 @@
 
 DEFINE_bool(move_files, false,
             "Move the SST files instead of copy when ingest into dataset");
+DEFINE_uint32(readahead_size, 512 * 1024,
+              "Using a large readahead size can typically improve the performance "
+              "of forward iteration");
 
 namespace nebula {
 namespace kvstore {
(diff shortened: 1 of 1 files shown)


src/kvstore/RocksEngine.cpp
```diff
@@ -15,6 +15,9 @@
 
 DEFINE_bool(move_files, false,
             "Move the SST files instead of copy when ingest into dataset");
+DEFINE_uint32(readahead_size, 512 * 1024,
```
critical27, line 18: Maybe just use the default value of rocksdb, it has a default value of 256KB. see https://github.com/facebook/rocksdb/pull/3282. Also it won't help a lot, our bottleneck is the `Seek` operation.

darionyaphet: We need a benchmark here

critical27: You could give a benchmark here. It only works for hdd. Block in ssd is not consecutive, so prefetch won't help.

bright-starry-sky: > You could give a benchmark here. It only works for hdd. Block in ssd is not consecutive, so prefetch won't help.

I agree with @critical27 , This parameter has an advantage for HDDs, but has a performance impact on SSDs because it has other overhead when turned on.

darionyaphet: ok  I'll close it firstly

PR #468, opened 2021-05-25
Feature/expand with dst

Shylock-Hg:
Expand to specific dest vertex.

Require https://github.com/vesoft-inc/nebula-common/pull/543


src/storage/exec/StorageIterator.h
```diff
@@ -114,6 +116,16 @@ class SingleEdgeIterator : public StorageIterator {
             return false;
         }
 
+        // check dst
+        if (dstId_.hasValue()) {
+            auto rawDstId = NebulaKeyUtils::getDstId(planContext_->vIdLen_, iter_->key());
+            // TODO optimize the compare
+            dstId_.value().resize(rawDstId.size());
+            if (dstId_.value() != rawDstId) {
```
critical27, line 124: Why not just push down a filter?

Shylock-Hg: So what's the filter expression?

src/storage/query/GetNeighborsProcessor.cpp
```diff
@@ -61,39 +61,71 @@ void GetNeighborsProcessor::doProcess(const cpp2::GetNeighborsRequest& req) {
...
-            auto vId = row.values[0].getStr();
-
-            if (!NebulaKeyUtils::isValidVidLen(spaceVidLen_, vId)) {
-                LOG(ERROR) << "Space " << spaceId_ << ", vertex length invalid, "
-                           << " space vid len: " << spaceVidLen_ << ",  vid is " << vId;
-                pushResultCode(nebula::cpp2::ErrorCode::E_INVALID_VID, partId);
-                onFinished();
-                return;
-            }
+    if (!withDst_) {
```
critical27, line 64: If we don't implement this feature by push down the filter. Maybe you could save a dstId in `PlanContext` or `EdgeContext`, and just pass it to `StorageIterator`, the template seems unnecessary.

But I sill prefer by adding a filter.

Shylock-Hg: There are pair of <src,dst> for each edge scan, so I think it should passed by the StoragePlan::go method.

src/storage/exec/StorageIterator.h
critical27: `edge._dst == xxx`?

Shylock-Hg: But the `xxx` is different for each vertex.It's not constant.

critical27: I know exactly what you mean here. We need to design related query grammar. What if we need to specify rank as well, add one extra template? It seems unacceptable to me.

Shylock-Hg: If you don't want the template parameter, I think could let the StoragePlan::go accept the Row type as the input.

PR #471, opened 2021-05-28
Fix add host overlap with another zone's host

darionyaphet:


critical27 (approved):
LGTM

panda-sheep (approved):

darionyaphet:
fix the meta client test case and please go on ~

src/meta/processors/zoneMan/UpdateZoneProcessor.cpp
```diff
@@ -40,14 +40,31 @@ void AddHostIntoZoneProcessor::process(const cpp2::AddHostIntoZoneReq& req) {
...
-    if (iter != hosts.end()) {
-        LOG(ERROR) << "Host " << host << " already exist in the zone " << zoneName;
-        handleErrorCode(nebula::cpp2::ErrorCode::E_EXISTED);
+    // check this host not exist in all zones
+    const auto& prefix = MetaServiceUtils::zonePrefix();
+    auto iterRet = doPrefix(prefix);
+    if (!nebula::ok(iterRet)) {
+        auto retCode = nebula::error(iterRet);
+        LOG(ERROR) << "Get zones failed, error: " << apache::thrift::util::enumNameSafe(retCode);
+        handleErrorCode(retCode);
```
bright-starry-sky, line 49: A suggestion, I think the line 16 - 39 are duplicate code. you can check it via zonePrefix. 
In addition, the code is confusing, for example getZoneId->doGet(indexKey) at line 17, and line 27 redo the doGet.
So I think you can use prefix instead of doGet for zone check. WDYT?

darionyaphet: first of all it will check the zone is existed make sure we can update it.

```
getZoneId->doGet(indexKey) at line 17, and line 27 redo the doGet.
```

what is it?

bright-starry-sky: > first of all it will check the zone is existed make sure we can update it.
> 
> Do you mean to check whether the zone exists when scanning?

you are right, my mistake. I confused the difference between `indexZoneKey` and `zoneKey`. I'm sorry I startled you!

bright-starry-sky (approved):
LGTM

panda-sheep (approved):
Good

PR #473, opened 2021-05-31
Shrink sizeof Value struct

dutor:
Depends on vesoft-inc/nebula-common#549 
Required by vesoft-inc/nebula-graph#1094

### DIFF
--- a/src/codec/RowWriterV2.cpp
+++ b/src/codec/RowWriterV2.cpp
@@ -780,21 +780,28 @@ WriteResult RowWriterV2::write(ssize_t index, const Time& v) noexcept {
 WriteResult RowWriterV2::write(ssize_t index, const DateTime& v) noexcept {
     auto field = schema_->field(index);
     auto offset = headerLen_ + numNullBytes_ + field->offset();
+    int16_t year = v.year;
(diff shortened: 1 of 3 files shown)


critical27 (approved):
LGTM

src/codec/RowWriterV2.cpp
```diff
@@ -780,21 +780,28 @@ WriteResult RowWriterV2::write(ssize_t index, const Time& v) noexcept {
 WriteResult RowWriterV2::write(ssize_t index, const DateTime& v) noexcept {
     auto field = schema_->field(index);
     auto offset = headerLen_ + numNullBytes_ + field->offset();
+    int16_t year = v.year;
```
Shylock-Hg, line 783: Maybe we could store the bit field directly in next incompatible version.

Shylock-Hg (approved):

src/codec/RowWriterV2.cpp
dutor: It might have trouble with the secondary index, but we could have a try.

bright-starry-sky (approved):
LGTM

PR #479, opened 2021-06-02
reuse newpart

panda-sheep:
As title

1. reuse newPart
2. fix maybe wait forever becase counter !=0


panda-sheep:
I’m sorry, through the conflict I found that pr 461 has fixed，so close.

PR #483, opened 2021-06-10
return empty when the TagName OR edgeName of current row is different…

nevermore3:
… from the TagName OR edgeName to be processed


src/storage/context/StorageExpressionContext.cpp
```diff
@@ -92,7 +92,7 @@ Value StorageExpressionContext::getSrcProp(const std::string& tagName,
                                            const std::string& prop) const {
     if (!isEdge_ && reader_ != nullptr) {
         if (tagName != name_) {
-            return Value::kNullValue;
+            return Value::kEmpty;
```
critical27, line 95: Line 114 perhaps need to change it as well.

critical27 (commented):
What is the conclusion about `NULL` and `EMPTY`?

nevermore3:
> What is the conclusion about `NULL` and `EMPTY`?

From the perspective of the user:  
```
Return EMPTY when the value does not exist
Return NULL when the value is NULL
```

From the perspective of the developer:
```
Return EMPTY when the value does not exist
Return NULL when the value is NULL OR EXCEPTION
```

Try to make the smallest changes to the semantics，Before a Release version。

src/storage/context/StorageExpressionContext.cpp
nevermore3: done

czpmango:
> > What is the conclusion about `NULL` and `EMPTY`?
> 
> From the perspective of the user:
> 
> ```
> Return EMPTY when the value does not exist
> Return NULL when the value is NULL
> ```
> 
> From the perspective of the developer:
> 
> ```
> Return EMPTY when the value does not exist
> Return NULL when the value is NULL OR EXCEPTION
> ```
> 
> Try to make the smallest changes to the semantics，Before a Release version。

What does "EXCEPTION" mean? Won't "EXCEPTION" report an error？

nevermore3:
> > > What is the conclusion about `NULL` and `EMPTY`?
> > 
> > 
> > From the perspective of the user:
> > ```
> > Return EMPTY when the value does not exist
> > Return NULL when the value is NULL
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > From the perspective of the developer:
> > ```
> > Return EMPTY when the value does not exist
> > Return NULL when the value is NULL OR EXCEPTION
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Try to make the smallest changes to the semantics，Before a Release version。
> 
> What does "EXCEPTION" mean? Won't "EXCEPTION" report an error？

during the execution phase, the data is abnormal, 
For example, vid type is different from the schema, return value::kNullBadType

critical27 (approved):
Good job

critical27:
> > > > What is the conclusion about `NULL` and `EMPTY`?

OK, the rules LGTM

yixinglu (approved):
We maybe need not the `EMPTY` later, now LGTM.

PR #484, opened 2021-06-10
unify reset/cleanup usage in Listener/Part

critical27:
1. Remove useless `lastId_` and `lastTerm_` in Listener. @bright-starry-sky, you can use `leaderCommitId_` to check if listener has catch up with leader. (when `leaderCommitId_ == 0`, could return false)
2. Unify the usage of `reset` and `cleanup` in RaftPart/Part/Listener:
* `reset` will clean all things (wal, Id, info about state machine), it will call `cleanup`
* `cleanup` only cleans info about state machine

### DIFF
--- a/src/kvstore/Listener.cpp
+++ b/src/kvstore/Listener.cpp
@@ -115,17 +115,13 @@ bool Listener::preProcessLog(LogID logId,
 
 bool Listener::commitLogs(std::unique_ptr<LogIterator> iter) {
     LogID lastId = -1;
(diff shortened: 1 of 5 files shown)


bright-starry-sky (dismissed):
artimg

bright-starry-sky:
File Conflicting

panda-sheep (approved):

bright-starry-sky (approved):
LGTM

PR #486, opened 2021-06-11
fix report error when field type and value not match in update edge

panda-sheep:
As title.

When the field type and the value type do not match and cannot be converted (RowWriterV2::write reports an error), nothing was written before, and now an error is reported.


liuyu85cn (approved):
LGTM

bright-starry-sky (approved):
LGTM

PR #496, opened 2021-06-18
Check the vid when get vertex properties.

### LINKED ISSUE #491: Unexpected result of getProps
The getProps won't return row of vertex which don't contains the request tag.
e.g. in space nba
If we have 
```
'Spurs'
```
as the request vids.
And 
```
player.age
```
as the request properties.
We will get empty result even if the vid 'Spurs' is existed in space.

Shylock-Hg:
Close #491


src/storage/query/GetPropProcessor.cpp
```diff
@@ -186,7 +186,8 @@ nebula::cpp2::ErrorCode GetPropProcessor::buildTagContext(const cpp2::GetPropReq
     } else {
         // not use const reference because we need to modify it when all property need to return
         auto returnProps = std::move(*req.vertex_props_ref());
-        ret = handleVertexProps(returnProps);
+        auto lackTags = lackTag(returnProps);  // Check the vid
```
critical27, line 189: Could consider add a flag in `GetPropRequest`, for now, the `lockTags` will force **every request** to check if each tag data exists. Do we really to do that? At least, at a todo here.

Please check graph ci as well.

Shylock-Hg: The graph test is ok in my local running. And in fact, we don't when to set the flag which determine whether scan all tags.

critical27: >  And in fact, we don't know when to set the flag which determine whether scan all tags.

Why? As your example, `match (a:player)--(b)--(c) return c.age`. It is `c.age`, not `player.age. 

Also, `fetch prop on *` is different from `fetch prop on player`. Why we can't tell the difference.

Shylock-Hg: There is no property of vertex, we just have property of tag. So if only scan the `player` tag, some exist vertex will be ignored.

critical27: Why graph can't set the flag which determine whether scan all tags. **The query can tell the difference between `scan all tags` and `only one tag`**

Shylock-Hg: We don't know when to set the flag to be true. Because in graph, we don't known which tag is attached to vertex.
For example:
```
fetch prop on player 'xxx'
```
Some vertex has player tag, some don't. And graph don't know it.

critical27: `fetch prop on player 'xxx`` -> query has specify only tag `player`
`fetch prop on * 'xxx'` -> query has specify all tags

Why we don't know

Shylock-Hg: If we only scan the player tag, if it's not attached to the vertex, this don't return one row.

critical27: We could modify it to return the row. But this is not the reason to read all tags all the time.

critical27: Check your PR #160, in that PR you don't return the row, and it involves another read. I reverted in #331 after discuss. Why we back and forth again and again? That is why we need to set a interface which can be specified **explicitly**, no more default behavior.

Shylock-Hg: It's required by https://github.com/vesoft-inc/nebula-graph/pull/1137, in this match we only collect the property used, so if the request tag is not attached to the vertex, the vertex won't return one row. It's mistake in match.

Shylock-Hg: And you could check this pr https://github.com/vesoft-inc/nebula-graph/pull/1150, in this case, shall we return the row?

critical27: vesoft-inc/nebula-graph#1137, If match need to return a row even if there is no tag attached to it, I'm OK to return a row with empty. 

vesoft-inc/nebula-graph#1150 is another issue, return or not both LGTM.  The case you add seems to work by coincidence, try `FETCH PROP ON person "3" YIELD person.id`. I believe no row is returned. 

That is why I insist query engine need to tell the behavior to storage.

Shylock-Hg: This won't be included in next version. Ok, I will add a flag and set true only in match.

critical27: 👍

Shylock-Hg: Modified.

src/storage/query/QueryBaseProcessor.h
```diff
@@ -7,6 +7,7 @@
 #ifndef STORAGE_QUERY_QUERYBASEPROCESSOR_H_
 #define STORAGE_QUERY_QUERYBASEPROCESSOR_H_
 
+#include <vector>
```
bright-starry-sky, line 10: delete it.

Shylock-Hg: Done.

PR #501, opened 2021-06-22
use local id in the space range for tag edge index

panda-sheep:
As title
1. Tag, edge, tag index and edge index  use space range id.
2. Compatible with the existing old schema creating  previously.

### DIFF
--- a/src/meta/processors/BaseProcessor.inl
+++ b/src/meta/processors/BaseProcessor.inl
@@ -189,6 +189,74 @@ ErrorOr<nebula::cpp2::ErrorCode, int32_t> BaseProcessor<RESP>::autoIncrementId()
 
 
 template<typename RESP>
+ErrorOr<nebula::cpp2::ErrorCode, int32_t> BaseProcessor<RESP>::getAvailableGolbalId() {
+    // A read lock has been added before call
+    static const std::string kIdKey = "__id__";
+    int32_t id;
+    std::string val;
+    auto ret = kvstore_->get(kDefaultSpaceId, kDefaultPartId, kIdKey, &val);
+    if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+        if (ret != nebula::cpp2::ErrorCode::E_KEY_NOT_FOUND) {
+            return ret;
+        }
+        id = 1;
+    } else {
+        id = *reinterpret_cast<const int32_t*>(val.c_str()) + 1;
+    }
+
+    return id;
+}
+
+
+template<typename RESP>
+ErrorOr<nebula::cpp2::ErrorCode, int32_t>
+BaseProcessor<RESP>::autoIncrementIdInSpace(GraphSpaceID spaceId) {
+    folly::SharedMutex::WriteHolder wHolder(LockUtils::localIdLock());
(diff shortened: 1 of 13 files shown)


src/meta/processors/BaseProcessor.inl
```diff
@@ -188,6 +188,43 @@ ErrorOr<nebula::cpp2::ErrorCode, int32_t> BaseProcessor<RESP>::autoIncrementId()
...
+    folly::SharedMutex::WriteHolder holder(LockUtils::localIdLock());
+    auto localIdkey = MetaServiceUtils::localIdKey(spaceId);
+    int32_t id;
+    std::string val;
+    auto ret = kvstore_->get(kDefaultSpaceId, kDefaultPartId, localIdkey, &val);
+    if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+        if (ret != nebula::cpp2::ErrorCode::E_KEY_NOT_FOUND) {
+            return ret;
+        }
+        id = 1;
```
critical27, line 203: You can't set to 1. This is a bit more complicated when there is old data, you need to think about it carefully. Better add a ut to mock this case as well.

panda-sheep: When upgrading, when the current space has an old schema, 
you only need to find the maximum value of the tagId, edgetype, and indexid under the current space,
and the id is the maximum value +1
Then insert the key value into the meta.

This approach has changed very little.

panda-sheep: But this needs to be done in the meta upgrade tool.

critical27: `maximum value of old autoIncrementId +1` sounds ok. Meta is upgraded in-place, no extra tools. If you set to 1, the old data will be messed up.

panda-sheep: Yes, I mean in-place updated meta.
Just deal with it here.

bright-starry-sky:
I can't find where `autoIncrementIdInSpace` is used for `create tag | edge` .

bright-starry-sky:
> I can't find where `autoIncrementIdInSpace` is used for `create tag | edge` .

This PR only deals with indexes, right?

panda-sheep:
> > I can't find where `autoIncrementIdInSpace` is used for `create tag | edge` .
> 
> This PR only deals with indexes, right?

tag, edge, tag index. edge index are all processed
I will immediately add another test with an existing schema

bright-starry-sky:
> > > I can't find where `autoIncrementIdInSpace` is used for `create tag | edge` .
> > 
> > 
> > This PR only deals with indexes, right?
> 
> tag, edge, tag index. edge index are all processed
> I will immediately add another test with an existing schema

yes,  I saw what processed. good job.

src/meta/processors/BaseProcessor.inl
panda-sheep: In order to be compatible with the existing old schema and implementation convenience, when the local_id key of the current space does not exist, it is directly set to autoIncrementId + 1.  thx

panda-sheep:
Ready, please continue.

bright-starry-sky (dismissed):
LGTM

critical27:
LGTM, wait the version has been released.

PR #502, opened 2021-06-22
fix balance thread hang up

critical27:
If future's callback is invoked before the future is set. And in the callback, baton will wait forever.

### DIFF
--- a/src/kvstore/wal/FileBasedWal.cpp
+++ b/src/kvstore/wal/FileBasedWal.cpp
@@ -315,7 +315,6 @@ void FileBasedWal::closeCurrFile() {
     timebuf.modtime = currInfo_->mtime();
     timebuf.actime = currInfo_->mtime();
     VLOG(1) << "Close cur file " << currInfo_->path() << ", mtime: " << currInfo_->mtime();
-    CHECK_EQ(utime(currInfo_->path(), &timebuf), 0);
     currInfo_.reset();
 }
 
--- a/src/kvstore/wal/FileBasedWal.cpp
+++ b/src/kvstore/wal/FileBasedWal.cpp
@@ -315,6 +315,7 @@ void FileBasedWal::closeCurrFile() {
     timebuf.modtime = currInfo_->mtime();
     timebuf.actime = currInfo_->mtime();
     VLOG(1) << "Close cur file " << currInfo_->path() << ", mtime: " << currInfo_->mtime();
+    utime(currInfo_->path(), &timebuf);
     currInfo_.reset();
 }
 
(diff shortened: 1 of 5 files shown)


src/meta/processors/admin/AdminClient.cpp
```diff
@@ -207,23 +207,21 @@ folly::Future<Status> AdminClient::updateMeta(GraphSpaceID spaceId,
...
-                                         apache::thrift::util::enumNameSafe(code).c_str()));
-            }
+    kv_->asyncMultiPut(kDefaultSpaceId, kDefaultPartId, std::move(data),
+        [this, p = std::move(pro)] (nebula::cpp2::ErrorCode code) mutable {
+            folly::via(ioThreadPool_.get(), [code, p = std::move(p)]() mutable {
+                if (code == nebula::cpp2::ErrorCode::SUCCEEDED) {
+                    p.setValue(Status::OK());
+                } else {
+                    p.setValue(Status::Error("Access kv failed, code: %s",
+                                            apache::thrift::util::enumNameSafe(code).c_str()));
```
darionyaphet, line 222: sorry to say alignment.

critical27: 😂 my poor vscode, hold on

darionyaphet (approved):

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -43,8 +43,8 @@ void HBProcessor::process(const cpp2::HBReq& req) {
        }
     }
 
-    VLOG(3) << "Receive heartbeat from " << host
-            << ", role = " << apache::thrift::util::enumNameSafe(req.get_role());
+    LOG(INFO) << "Receive heartbeat from " << host
```
bright-starry-sky, line 46: VLOG（1）？

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -315,7 +315,7 @@ void FileBasedWal::closeCurrFile() {
     timebuf.modtime = currInfo_->mtime();
     timebuf.actime = currInfo_->mtime();
     VLOG(1) << "Close cur file " << currInfo_->path() << ", mtime: " << currInfo_->mtime();
-    CHECK_EQ(utime(currInfo_->path(), &timebuf), 0);
```
bright-starry-sky, line 318: what's reason? no need to check the system error?

bright-starry-sky (commented):
great job，It's a difficult problem to identify.

src/meta/processors/admin/HBProcessor.cpp
critical27: Previously was INFO, modified to VLOG(3) by others. It is the most convenient way to tell which is meta leader.

liuyu85cn (approved):

PR #503, opened 2021-06-24
Concurrently go/lookup/fetch

critical27:
Follow up of #498, this will work. 

Ready to review. Change list:
1. `PlanContext` save all info during process, and some of them would be changed after request is parsed. Since we need to run them in different thread, I separate fields unchanged into `PlanContext` and fields would be changed into `RunTimeContext`.
2. For each thread, create a `RunTimeContext` and all others which are not thread safe (ExpressionContext, Expression, DataSet to save result). When each thread finish, merge the result and then return.

Go/lookup/fetch in storage could support execute on multiple thread. The feature is disabled by default. Adding `--query_concurrently=true` in storage conf to enable it.


critical27:
It works for now. Todo: Expression safety between multiple threads. As far as I know, graphd has clone every expression among threads.

bright-starry-sky (approved):
LGTM

panda-sheep (approved):
Excellent!

PR #516, opened 2021-07-06
Re-test index write performance since the concurrentMap instead of asyncAtomicOp.

bright-starry-sky:
Add bench mark for index write. refer nebula 1.0.


critical27 (approved):

panda-sheep (approved):
Good ！

PR #519, opened 2021-07-08
Fix show stats

critical27:
Both BlockBasedTable and PlainTable will failed in some certain case.

1. Both will failed when use HashSkipList as memtable.
2. BlockBasedTable will fail when specify `prefix_same_as_start`, it is ok to remove it since we check it externally.
 
Will dig more later and add some test case.

### DIFF
--- a/src/kvstore/RocksEngine.cpp
+++ b/src/kvstore/RocksEngine.cpp
@@ -220,7 +220,6 @@ nebula::cpp2::ErrorCode
 RocksEngine::prefix(const std::string& prefix,
                     std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
-    options.prefix_same_as_start = true;
     rocksdb::Iterator* iter = db_->NewIterator(options);
     if (iter) {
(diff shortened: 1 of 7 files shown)


src/kvstore/RocksEngine.cpp
```diff
@@ -234,7 +233,6 @@ RocksEngine::rangeWithPrefix(const std::string& start,
                              const std::string& prefix,
                              std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
-    options.prefix_same_as_start = true;
```
panda-sheep, line 237: Good job，thx
Perhaps just removing `prefix_same_as_start` is enough. In this way rocksdb uses the default value of `prefix_same_as_start`.

liuyu85cn (approved):

bright-starry-sky (approved):
LGTM

critical27:
Hold on a sec, do some research.

critical27:
Ready....

bright-starry-sky (approved):

liuyu85cn (approved):

PR #517, opened 2021-07-06
Improved data write logic to avoid reading dirty data

bright-starry-sky:
We are using `ConcurrentHashMap` to solve the performance issues of concurrent writes.
The current write logic may cause dirty data to be read, for example:
Thread A : -----> read -> lock -> write
Thread B : read----------------------------> lock -> write
If thread A terminates first, it will cause thread B to read dirty data.

changed to ：
Thread A : lock -----> read -> write -> unlock
Thread B : lock - > read-----------------------------> write -> unlock

PR #523, opened 2021-07-12
Revert->Revert "Improved data write logic to avoid reading dirty data"

bright-starry-sky:
Reverts vesoft-inc/nebula-storage#517


bright-starry-sky:
Please hold this PR. One little change is required

PR #528, opened 2021-07-14
fix task in recovered balance plan is still marked as FAILED

critical27:
When task is failed for some reason, when recovering, they are marked as `IN_PROGRESS` again, but `balance data id` will still show `FAILED`. Just save the status when plan is recovered.

### DIFF
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -36,6 +36,10 @@ folly::Future<Status> AdminClient::transLeader(GraphSpaceID spaceId,
     if (it == peers.end()) {
         return Status::PartNotFound();
     }
+    if (peers.size() == 1 && peers.front() == leader) {
+        // if there is only one replica, skip transfer leader phase
+        return Status::OK();
+    }
     auto target = dst;
     if (dst == kRandomPeer) {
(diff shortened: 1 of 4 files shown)


panda-sheep (dismissed):
Well done！

critical27:
There will be more changes, hold on one second.

critical27:
close https://github.com/vesoft-inc/nebula-storage/issues/458
close https://github.com/vesoft-inc/nebula-storage/issues/458

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -36,6 +36,10 @@ folly::Future<Status> AdminClient::transLeader(GraphSpaceID spaceId,
     if (it == peers.end()) {
         return Status::PartNotFound();
     }
+    if (peers.size() == 1 && peers.front() == leader) {
```
darionyaphet, line 39: make sure the leader is being selected?

critical27: No, transfer leader will failed in some case when replica is 1. see https://github.com/vesoft-inc/nebula-storage/issues/458.

panda-sheep (approved):

liuyu85cn (approved):
LGTM

Issue #530, opened 2021-07-21
VertexCache will destroy the consistency of raft


cangfengzhs:
1. When reading data from VertexCache, it does not check whether the current part is the leader.
2. When the leader transLeader from part A to part B, and then transLeader to part A, the data in the cache and the data in kvstore may be different
![image](https://user-images.githubusercontent.com/22708345/126481363-78053924-fd28-4d17-b3a4-48cb8c961050.png)

3. When insert/update and read are executed concurrently, vertexCache may hold older data, which is different from that in kvstore.
![image](https://user-images.githubusercontent.com/22708345/126482963-fd6fd6eb-1fbc-4e11-bb73-35842da63433.png)

critical27:
Good job, long history, could replace it with row cache later.

PR #532, opened 2021-07-22
Replace vertex lru cache with rocksdb row cache

critical27:
1. Replace vertex LRU cache with rocksdb row cache, to avoid annoying inconsistency
2. Since no mvcc any more, replace some prefix with get, just like https://github.com/vesoft-inc/nebula/pull/2377 in 1.0

Close https://github.com/vesoft-inc/nebula-storage/issues/530.


src/storage/exec/EdgeNode.h
```diff
@@ -116,7 +86,28 @@ class FetchEdgeNode final : public EdgeNode<cpp2::EdgeKey> {
                   Expression* exp = nullptr)
         : EdgeNode(context, edgeContext, edgeType, props, expCtx, exp) {}
 
+    bool valid() const override {
+        return valid_;
+    }
+
+    void next() override {
+        valid_ = false;
```
bright-starry-sky, line 94: Why?

critical27: They are for fetch vertex/edge, only first data should be used. Previously this job is do in `StorageIterator`, when we change from `prefix` to `get`, there is no more rocksdb iterator.

cangfengzhs (approved):
LGTM

bright-starry-sky (approved):
LGTM

PR #541, opened 2021-08-04
Delete tag

critical27:
Rely on https://github.com/vesoft-inc/nebula-common/pull/603

### DIFF
--- a/src/storage/mutate/DeleteTagsProcessor.cpp
+++ b/src/storage/mutate/DeleteTagsProcessor.cpp
@@ -1,0 +1,162 @@
+/* Copyright (c) 2021 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "storage/StorageFlags.h"
+#include "storage/mutate/DeleteTagsProcessor.h"
+#include "utils/IndexKeyUtils.h"
+#include "utils/NebulaKeyUtils.h"
+#include "utils/OperationKeyUtils.h"
+
+namespace nebula {
+namespace storage {
+
+ProcessorCounters kDelTagsCounters;
+
+void DeleteTagsProcessor::process(const cpp2::DeleteTagsRequest& req) {
+    spaceId_ = req.get_space_id();
+    const auto& parts = req.get_parts();
+
+    CHECK_NOTNULL(env_->schemaMan_);
+    auto ret = env_->schemaMan_->getSpaceVidLen(spaceId_);
+    if (!ret.ok()) {
+        LOG(ERROR) << ret.status();
+        for (auto& part : parts) {
+            pushResultCode(nebula::cpp2::ErrorCode::E_INVALID_SPACEVIDLEN, part.first);
+        }
+        onFinished();
+        return;
+    }
+    spaceVidLen_ = ret.value();
+    callingNum_ = parts.size();
+
+    CHECK_NOTNULL(env_->indexMan_);
+    auto iRet = env_->indexMan_->getTagIndexes(spaceId_);
+    if (!iRet.ok()) {
+        LOG(ERROR) << iRet.status();
+        for (auto& part : parts) {
+            pushResultCode(nebula::cpp2::ErrorCode::E_SPACE_NOT_FOUND, part.first);
+        }
+        onFinished();
+        return;
+    }
+    indexes_ = std::move(iRet).value();
+
+    CHECK_NOTNULL(env_->kvstore_);
+    if (indexes_.empty()) {
+        std::vector<std::string> keys;
+        keys.reserve(32);
+        for (const auto& part : parts) {
+            auto partId = part.first;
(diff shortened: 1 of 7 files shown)


src/storage/mutate/DeleteTagsProcessor.cpp
```diff
@@ -0,0 +1,162 @@
...
+        keys.reserve(32);
+        for (const auto& part : parts) {
+            auto partId = part.first;
+            const auto& delTags = part.second;
+            keys.clear();
+            auto code = nebula::cpp2::ErrorCode::SUCCEEDED;
+            for (const auto& entry : delTags) {
+                const auto& vId = entry.get_id().getStr();
+                for (const auto& tagId : entry.get_tags()) {
+                    auto key = NebulaKeyUtils::vertexKey(spaceVidLen_, partId, vId, tagId);
```
Shylock-Hg, line 59: Shall we delete edge when delete last tag of vertex?

critical27: We could consider later, this involves how we deal with dangle edges.

Shylock-Hg: I think we should keep same with delete vertex but it's difficult to do. So maybe could let delete tag can't delete the last tag.

src/storage/mutate/DeleteTagsProcessor.cpp
```diff
@@ -0,0 +1,162 @@
...
+    auto ret = env_->schemaMan_->getSpaceVidLen(spaceId_);
+    if (!ret.ok()) {
+        LOG(ERROR) << ret.status();
+        for (auto& part : parts) {
+            pushResultCode(nebula::cpp2::ErrorCode::E_INVALID_SPACEVIDLEN, part.first);
+        }
+        onFinished();
+        return;
+    }
+    spaceVidLen_ = ret.value();
```
bright-starry-sky, line 32: is better `std::move(ret).value `?

src/storage/mutate/DeleteTagsProcessor.cpp
```diff
@@ -0,0 +1,162 @@
...
+                const auto& vId = entry.get_id().getStr();
+                for (const auto& tagId : entry.get_tags()) {
+                    auto key = NebulaKeyUtils::vertexKey(spaceVidLen_, partId, vId, tagId);
+                    keys.emplace_back(std::move(key));
+                }
+            }
+            if (code != nebula::cpp2::ErrorCode::SUCCEEDED) {
+                handleAsync(spaceId_, partId, code);
+                continue;
+            }
```
bright-starry-sky, line 66: Do we need to check the code here?

src/storage/mutate/DeleteTagsProcessor.cpp
```diff
@@ -0,0 +1,162 @@
...
+                    const auto& cols = index->get_fields();
+                    auto valuesRet = IndexKeyUtils::collectIndexValues(reader.get(), cols);
+                    if (!valuesRet.ok()) {
+                        return nebula::cpp2::ErrorCode::E_INVALID_DATA;
+                    }
+                    auto indexKey = IndexKeyUtils::vertexIndexKey(spaceVidLen_,
+                                                                    partId,
+                                                                    indexId,
+                                                                    vId,
+                                                                    std::move(valuesRet).value());
```
bright-starry-sky, line 141: alignment

bright-starry-sky (commented):
Good job, the overall logic is fine for me.

src/storage/mutate/DeleteTagsProcessor.cpp
critical27: Why delete tag need to keep same with delete vertex? We need to read once for each tag? Besides, how can we make sure the tag we are deleting is the last one?

Delete vertex need to be handled with txn.

src/storage/mutate/DeleteTagsProcessor.cpp
critical27: Ah, dirty code, let me remove them

src/storage/mutate/DeleteTagsProcessor.cpp
critical27: Not a big difference as for integer?

bright-starry-sky (approved):
LGTM，nice job.

src/storage/mutate/DeleteTagsProcessor.cpp
Shylock-Hg: We could add a special tag with id 0, in this case delete tag won't delete vertex.

panda-sheep (approved):
Woohoo!

src/storage/mutate/DeleteTagsProcessor.cpp
critical27: This is not just the delete tag's problem. Since we allow dangling edges, we only handle `delete last tag -> delete vertex` is meaningless.

liuyu85cn (approved):
LGTM

PR #544, opened 2021-08-05
Re-change the options for rebuild index batch number

bright-starry-sky:
I'm so sorry. the configuration file was wrong in the previous PR #540. Let's correct that in this PR.
:sweat_smile:

### DIFF
--- a/src/storage/StorageFlags.cpp
+++ b/src/storage/StorageFlags.cpp
@@ -26,6 +26,9 @@ DEFINE_int32(rebuild_index_batch_num, 256,
 DEFINE_int32(rebuild_index_locked_threshold, 256,
              "The locked threshold will refuse writing.");
 
+DEFINE_int32(rebuild_index_process_interval, 30000,
+             "Index rebuild processing interval, microsecond.");
+
 DEFINE_bool(enable_vertex_cache, true, "Enable vertex cache");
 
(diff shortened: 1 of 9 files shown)


randomJoe211 (dismissed):

critical27:
Hold on for one sec.

bright-starry-sky:
I discussed it with @critical27 , `rebuild_index_batch_num` is not very effective as the speed limit. added the new option `rebuild_index_process_interval` for speed limit.

src/storage/StorageFlags.cpp
```diff
@@ -26,6 +26,9 @@ DEFINE_int32(rebuild_index_batch_num, 256,
 DEFINE_int32(rebuild_index_locked_threshold, 256,
              "The locked threshold will refuse writing.");
 
+DEFINE_int32(rebuild_index_process_interval, 30000,
```
critical27, line 29: 3ms is enough I think? 30ms means there are 33 request at most for one second.

bright-starry-sky: > 3ms is enough I think? 30ms means there are 33 request at most for one second.

good catch. fixed.

critical27 (approved):
LGTM

darionyaphet (approved):

PR #545, opened 2021-08-06
Add interface to register callback when raft part elected

liuyu85cn:


liuyu85cn:
CLA issue, close this. open #546 instead

PR #548, opened 2021-08-10
get meta version

critical27:
Return meta version in hb resp.

### DIFF
--- a/src/meta/processors/admin/HBProcessor.cpp
+++ b/src/meta/processors/admin/HBProcessor.cpp
@@ -9,13 +9,14 @@
 #include "common/time/WallClock.h"
 #include "meta/ActiveHostsMan.h"
 #include "meta/KVBasedClusterIdMan.h"
-
-DEFINE_bool(hosts_whitelist_enabled, false, "Check host whether in whitelist when received hb");
+#include "meta/MetaVersionMan.h"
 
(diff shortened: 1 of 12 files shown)


src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -82,6 +71,14 @@ void HBProcessor::process(const cpp2::HBReq& req) {
     } else if (nebula::error(lastUpdateTimeRet) == nebula::cpp2::ErrorCode::E_KEY_NOT_FOUND) {
         resp_.set_last_update_time_in_ms(0);
     }
+
+    auto version = metaVersion_.load();
+    if (version == -1) {
+        metaVersion_.store(static_cast<int64_t>(MetaVersionMan::getMetaVersionFromKV(kvstore_)));
```
Shylock-Hg, line 77: I think you could initialize it when metad startup.

critical27: Yeah, when metad startup, it has already set this version. Just to avoid use a global.

bright-starry-sky (approved):

liuyu85cn (approved):

critical27:
Will open another PR to merge into 2.5.0

